PR #7, opened 2023-01-19
Initial commit of hipscat-import

delucchi-cmu:
This is mostly a port of code already reviewed under astronomy-commons/lsd2/src/partitioner.

Main differences include:
- Removing notion of a "single runner" that doesn't depend on dask. This just goes all-in on dask, and reduces abstraction layers.
- Treats all data to be mapped as a dask dataframe, instead of a pandas dataframe (useful for larger parquet file inputs, to avoid dask out of memory issues).
- A lot of the interesting pixel math and the catalog-level metadata files are handled by the new `hipscat` package, which is added as a dependency here.


src/hipscat_import/map_reduce.py
```diff
@@ -0,0 +1,138 @@
+"""Import a set of non-hipscat files using dask for parallelization"""
+
+import os
+
+import dask.dataframe as dd
```
maxwest-uw, line 5: why do we have two separate files name `dask_map_reduce` and `map_reduce` that both use dask? I feel like that might cause some complaints if a user is trying to avoid using dask 😅 maybe the first one should be `parallel_map_reduce` or something like that?

src/hipscat_import/dask_map_reduce.py
```diff
@@ -0,0 +1,99 @@
...
+import hipscat.io.write_metadata as io
+import hipscat.pixel_math as pixel_math
+import numpy as np
+from dask.distributed import Client, progress, wait
+
+import hipscat_import.map_reduce as mr
+from hipscat_import.arguments import ImportArguments
+
+
+def _generate_histogram(args, client):
```
maxwest-uw, line 12: this could be a future task to replace and I see that this one is parallelized, but is this histogram method redundant with the one found in the `hipscat` package?

maxwest-uw (commented):
had two questions, but overall everything is looking good!! starting to feel a lot more organized than all our prototypes haha

src/hipscat_import/dask_map_reduce.py
delucchi-cmu: This one should really be _map, since it's doing more than just generating a histogram of object counts (also makes it parallel with the below _reduce)

The one in hipscat is *only* getting the counts of objects per-pixel.

src/hipscat_import/map_reduce.py
delucchi-cmu: Fair point. This was an attempt to keep the dask-y pipeline worker setup distinct from the interesting logic, mostly for testing and cleanliness. Since the whole pipeline relies on dask, the name of the file is kinda silly.

maxwest-uw (approved):

PR #8, opened 2023-01-30
Use copier to keep up with best practices

delucchi-cmu:
Use the LINCC Frameworks project template at https://github.com/lincc-frameworks/python-project-template

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,133 @@
-## Byte-compiled, execution cache
+# Byte-compiled / optimized / DLL files
 __pycache__/
-dask-worker-space
-.coverage*
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,14 +1,133 @@
-## Byte-compiled, execution cache
+# Byte-compiled / optimized / DLL files
 __pycache__/
-dask-worker-space
```
drewoldag, line 3: Looks like you'll need to add `dask-worker-space` back to the .git ignore file.

drewoldag (approved):
This looks good too. Just to be clear, the workflow file `python-package.yml` is in a very draft state, but will confirm that a project at least builds. As far as why the build is failing when trying to find the `hipscat` dependency, is a bit of mystery though.

PR #10, opened 2023-02-01
Better handle type promotion during map reduce

delucchi-cmu:
- Stop using dask distributed dataframes for data loading - running into too many extra memory and type coercion problems to be worth it.
- Optionally provide a schema file to use for data types for loading a csv
- Use pandas when combining multiple parquet files (handles null->non-null promotion under mixed type circumstances better than raw pyarrow)
- Also reduces some indentation levels in a couple tests.


src/hipscat_import/arguments.py
```diff
@@ -233,6 +234,7 @@ def from_params(
         pixel_threshold=1_000_000,
         debug_stats_only=False,
         filter_function=None,
+        schema_file=None,
```
smcguire-cmu, line 237: Should the schema_file input be added to the `from_command_line` method as well?

src/hipscat_import/map_reduce.py
```diff
@@ -49,14 +37,17 @@ def map_to_pixels(
...
-        data = dd.from_pandas(
-            Table.read(input_file, format="fits").to_pandas(), chunksize=500_000
-        )
+        data = Table.read(input_file, format="fits").to_pandas()
     elif file_format == "parquet":
-        data = dd.read_parquet(input_file, engine="pyarrow")
+        data = pd.read_parquet(input_file, engine="pyarrow")
+        data.reset_index(inplace=True)
     else:
         raise NotImplementedError(f"File Format: {file_format} not supported")
```
smcguire-cmu, line 52: If we want to add more formats in the future, this could probably be refactored (e.g. using a factory instead of all the if statements), but it's fine for now

tests/hipscat_import/test_run_import.py
```diff
@@ -45,39 +45,39 @@ def test_dask_runner():
...
+        # Check that the catalog parquet file exists and contains correct object IDs
+        output_file = os.path.join(
+            args.catalog_path, "Norder0/Npix11", "catalog.parquet"
+        )
+
+        expected_ids = [*range(700, 831)]
+        ft.assert_parquet_file_ids(output_file, "id", expected_ids)
 
 
 @pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
smcguire-cmu, line 83: What's the reason we're ignoring this?

delucchi-cmu: It's happening somewhere inside dask, and there's nothing we can do about it.

smcguire-cmu (approved):

PR #11, opened 2023-02-01
Change dev dependencies name inline with project template

smcguire-cmu:
Change the name of the dev dependencies to match the project templates and other projects, similar to astronomy-commons/hipscat#13.

Also updated the README to include instructions for installing with dev dependencies on Mac.


delucchi-cmu (approved):
Thanks for the cleanup!

PR #18, opened 2023-02-06
Update README.md

maxwest-uw:
when installing the dev dependencies we especially want to include the `-e` to make sure that the package is editable (this caused me five minutes of stress that I'd prefer to save everyone else from 😛 )


smcguire-cmu (approved):
Yes, you're right, sorry about that!

PR #26, opened 2023-02-16
Prefactor arguments.

delucchi-cmu:
- separate command line parsing source.
- initialize from constructor (wild!)


jeremykubica:
Overall this looks good. One quick question: What is the motivation behind moving the command line parsing to a new file. It seems like it could have continued to live in the same file. Since this is a pre-factoring, I am guessing there is some down stream motivation.

delucchi-cmu:
Primarily, the motivation is to shrink the file. The functionality is so separate that the unit tests are already split between two files, so this change mirrors that.

My next move is to add more arguments for CSV parsing, and it was feeling unwieldy.

jeremykubica:
Thanks. That makes sense.

jeremykubica (approved):

PR #28, opened 2023-02-16
Read files using a generative file reader

delucchi-cmu:
There are too many options for reading a CSV that a user may want to specify, so this allows expert users to fully specify the way in which a file should be read.

This is re-structured to support effectively an iterator over the file, which is useful when you have a file that cannot/should not be read into memory all at once.


tests/hipscat_import/test_map_reduce.py
```diff
@@ -149,12 +147,47 @@ def test_map_small_sky_order0(small_sky_single_file):
...
 
-@pytest.mark.filterwarnings("ignore::DeprecationWarning")
+def test_map_small_sky_chunk1(small_sky_single_file):
+    """Test loading the small sky catalog and partitioning each object one row at a time"""
+
+    def csv_single_line_reader(input_file):
+        with pd.read_csv(input_file, chunksize=1) as reader:
+            for chunk in reader:
+                chunk.reset_index(inplace=True)
+                yield chunk
```
smcguire-cmu, line 162: Another instance where a reader class could have all this logic with a chunksize argument instead of writing out the function

src/hipscat_import/file_readers.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_file_reader(file_format):
+    """Get a generator file reader for common file types"""
+    if "csv" in file_format:
+        return csv_reader
+    if file_format == "fits":
+        return fits_reader
+    if file_format == "parquet":
+        return parquet_reader
```
smcguire-cmu, line 16: I think this could be refactored further. We could have a [factory method](https://www.geeksforgeeks.org/factory-method-python-design-patterns/) to generate the correct file_reader from the file_format string.

smcguire-cmu (commented):
I think it's not weird at all and improves the file reading code a lot

src/hipscat_import/arguments.py
```diff
@@ -27,7 +29,7 @@ def __init__(
         pixel_threshold=1_000_000,
         debug_stats_only=False,
         filter_function=None,
-        schema_file=None,
```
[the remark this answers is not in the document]
smcguire-cmu: Ah cool that makes sense

src/hipscat_import/file_readers.py
```diff
@@ -0,0 +1,104 @@
...
+            header=header,
+            schema_file=schema_file,
+            column_names=column_names,
+            type_map=type_map,
+            separator=separator,
+        ).read
+    if file_format == "fits":
+        return fits_reader
+    if file_format == "parquet":
+        return ParquetReader(chunksize=chunksize).read
```
smcguire-cmu, line 33: I was thinking the argument could take the object rather than just the method of the object. Could be useful in case we need any extra functionality from the file readers in the future but this works

delucchi-cmu: Sounds like a good idea, but also sounds like it's safe to wait until we see more need.

src/hipscat_import/file_readers.py
delucchi-cmu: My reading of the factory as linked is that you'd instantiate all the types and only call their methods when the appropriate arguments come up. In this case, we'd only want to instantiate one kind of reader for the whole execution of the pipeline, so the setup of the factory would be unnecessary overhead. Happy to talk further (or punt on more changes for now).

smcguire-cmu: I think that example instantiates all the types because that's what they want, but the factory itself just instantiates a single type. And it does so by looking up a string into a dictionary instead of if statements, which is neater and easier to add more formats to, including that you can do it programmatically. (If you really wanted to get fancy, you could use a decorator to add new file formats without having to change the factory class at all)

smcguire-cmu: But yeah, it's not necessary right now if we just want to get this merged

smcguire-cmu (approved):

PR #31, opened 2023-02-22
Implement pipeline resume

delucchi-cmu:
- Log completed stages to intermediate results
- Skip completed stages if resuming from intermediate results
- Explicitly remove files after successful pipeline execution


src/hipscat_import/__init__.py
```diff
@@ -4,4 +4,17 @@
 from .command_line_arguments import parse_command_line
 from .file_readers import CsvReader, ParquetReader, fits_reader, get_file_reader
 from .map_reduce import map_to_pixels, reduce_pixel_shards
+from .resume_files import (
+    clean_resume_files,
```
jeremykubica, line 8: Any reason not to just import * at this point? I know there are a few private type functions in those files, but you are already using the _ prefix to indicate those should be treated as internal only.

src/hipscat_import/arguments.py
```diff
@@ -115,44 +114,52 @@ def _check_paths(self):
...
-        self.contexts.append(tmp_dir)
+            self.tmp_path = os.path.join(self.catalog_path, "intermediate")
+        if not self.resume:
+            existing_tmp_files = glob.glob(f"{self.tmp_path}/*")
+            if existing_tmp_files:
+                raise ValueError(
+                    f"tmp_path ({self.tmp_path}) contains intermediate files."
+                    " choose a different directory or use --resume flag"
+                )
+        os.makedirs(self.tmp_path, exist_ok=True)
```
jeremykubica, line 151: I want to make sure that I understand this change of behavior. Moving from the `tempfile.TemporaryDirectory` to `os.makedirs` means the temporary file directory persists if the job is killed. The advantage of that the intermediate files exist for a resume. The disadvantage is that you now need to clean up. Is that all correct?

src/hipscat_import/resume_files.py
```diff
@@ -0,0 +1,96 @@
...
+
+def read_histogram(tmp_path, highest_healpix_order):
+    """Read a numpy array at the indicated directory.
+    Otherwise, return histogram of appropriate shape."""
+    if os.path.exists(os.path.join(tmp_path, "mapping_histogram.csv")):
+        return np.loadtxt(
+            os.path.join(tmp_path, "mapping_histogram.csv"),
+            dtype=np.ulonglong,
+            delimiter=",",
+        )
```
smcguire-cmu, line 21: Could we use a binary file instead of text for this? Performance and storage might be better. [Some options compared here.](https://stackoverflow.com/questions/9619199/best-way-to-preserve-numpy-arrays-on-disk)

src/hipscat_import/run_import.py
```diff
@@ -1,54 +1,76 @@
...
+    for future, result in tqdm(
         as_completed(futures, with_results=True),
         desc="Mapping  ",
         total=len(futures),
         disable=(not args.progress_bar),
     ):
         raw_histogram = np.add(raw_histogram, result)
+        rf.write_histogram(args.tmp_path, raw_histogram)
+        rf.write_mapping_key(args.tmp_path, future.key)
+    rf.set_mapping_done(args.tmp_path)
```
jeremykubica, line 52: What happens if the job crashes between lines 51 and 52? The histograms and mapping keys have been written out, but the mapping file handle hasn't been set. I realize this is a very low likelihood probability, but thought I would ask. Would this end up reprocessing those entries and double counting?

src/hipscat_import/run_import.py
```diff
@@ -1,54 +1,76 @@
...
 
 import hipscat_import.map_reduce as mr
+import hipscat_import.resume_files as rf
 from hipscat_import.arguments import ImportArguments
 
 
 def _map_pixels(args, client):
     """Generate a raw histogram of object counts in each healpix pixel"""
 
+    raw_histogram = rf.read_histogram(args.tmp_path, args.highest_healpix_order)
```
smcguire-cmu, line 18: I think using the non shortened import might make this more readable, or shortening to something like `resume` instead of `rf`

src/hipscat_import/command_line_arguments.py
```diff
@@ -123,6 +123,17 @@ def parse_command_line(cl_args):
...
+        "--resume",
+        help="if set, the pipeline will try to resume from a previous failed pipeline progress",
+        action="store_true",
+    )
+    group.add_argument(
+        "--no_resume",
+        help="if set, the pipeline will exit if existing intermediate files are found",
+        dest="resume",
+        action="store_false",
+    )
```
jeremykubica, line 136: Is it worth having an explicit "clean_resume" flag? If there are intermediate files the job will either continue or exit (which makes sense). In the case it exits, I think the user needs to clean the files manually. Adding a path to do that automatically could save a step in the future.

src/hipscat_import/run_import.py
```diff
@@ -1,54 +1,76 @@
...
+    if rf.is_mapping_done(args.tmp_path):
+        return raw_histogram
+
+    mapped_paths = rf.read_mapping_keys(args.tmp_path)
+
     futures = []
     for i, file_path in enumerate(args.input_paths):
+        if file_path in mapped_paths:
+            continue
```
smcguire-cmu, line 27: After practicing lots of interview questions I feel I have to point out mapped_paths could be made into a set to make this check quicker but it's probably a pointless optimization

src/hipscat_import/resume_files.py
```diff
@@ -0,0 +1,96 @@
...
+    """Read a resume log file, containing timestamp and keys."""
+    if os.path.exists(file_name):
+        mapping_log = pd.read_csv(
+            file_name,
+            delimiter="\t",
+            header=None,
+            names=["time", "key"],
+        )
+        return mapping_log["key"].values
+    return []
```
smcguire-cmu, line 75: `.values` returns a numpy array, empty path returns a list. Could cause some issues if used expecting one data type.

src/hipscat_import/run_import.py
smcguire-cmu: I think if the mapping key has been written then the next time it is run that file would be skipped from processing in lines 26 and 27

delucchi-cmu: Yeah - I don't know of any clever way to make these two operations atomic. I think we could even get killed mid-write and only part of the line would be available (or some subset of the histogram file would be updated), so combining into a single file doesn't resolve the problem.

This *would* cause a failure in the *reduce* stage, in a weird way:
- say input 5 has 12 objects in pixel 7
- if the job is killed between these lines, the histogram will have 12 objects in pixel 7, but the mapping key won't be present
- on resume, we'll still process input 5. we'll only overwrite the shard files generated on the first pass, and not write *new* ones.
- we'll add another 12 objects to pixel 7
- when we reduce, we set an expectation of finding 24 objects at pixel 7, but only 12 are written
- we raise an error inside reduce pixels

If we switch the order of these two lines and the job is killed between the steps:
- the histogram will have 0 objects in pixel 7, and the mapping key will be present
- on resume, we won't process input 5, so the histogram will still have 0 objects in pixel 7.
- when we reduce, we'll find 12 objects, expecting 0, and raise an error.

I'm open to suggestions!

jeremykubica: Warning: I don't have a **good** suggestion. I was wondering if you should include another file-based indicator (called start) that you write before the writes begin. When you are resuming you can skip anything that has "start" and "done", do anything that has neither, and fail if it only has one (or delete the intermediate files and restart).

jeremykubica: To be clear that wouldn't cover every case. I also don't have a way to make it fully atomic.

jeremykubica: All this might be over-engineering for a very rare case though.

smcguire-cmu: @delucchi-cmu is this is a problem about the operations on line 50 and 51 (Writing the histogram and writing the making key) or on line 51 and 52 (writing the histogram and keys and writing the done file)?

smcguire-cmu: If it's the former, another start writing log file as Jeremy suggests could work, if it's the latter, I'm not sure I understand the issue, sorry.

src/hipscat_import/resume_files.py
delucchi-cmu: Yup!

src/hipscat_import/run_import.py
delucchi-cmu: Ah! I misunderstood the original question. If we fail between 51 and 52, then that's totally fine. When we pick back up to resume
- we'll load the `mapping_histogram`, which will contain the full histogram at that point
- we'll load the `mapping_log`, which will contain all the input files
- as we loop through the input files, we'll just skip all of them
- at the end of the loop, we'll write out the done file
- if we need to resume again after that, the `mapping_done` file will be around and everything will be peachy.

Adding extra guard around the histogram writing so we can find out quickly that the intermediate files have been corrupted.

src/hipscat_import/run_import.py
delucchi-cmu: Done.

src/hipscat_import/run_import.py
delucchi-cmu: It's also a 4 character change, and the input list should already be sorted. So seems like it won't be expensive to turn into a set.

src/hipscat_import/run_import.py
```diff
@@ -1,54 +1,77 @@
...
     return raw_histogram
 
 
 def _reduce_pixels(args, destination_pixel_map, client):
     """Loop over destination pixels and merge into parquet files"""
 
+    if resume.is_reducing_done(args.tmp_path):
+        return
+
+    reduced_keys = resume.read_reducing_keys(args.tmp_path)
```
smcguire-cmu, line 63: Same set optimization could be made here

src/hipscat_import/resume_files.py
smcguire-cmu: Awesome, thanks!

src/hipscat_import/resume_files.py
```diff
@@ -0,0 +1,104 @@
...
+    if os.path.exists(file_name):
+        with open(file_name, "rb") as file_handle:
+            return frombuffer(file_handle.read(), dtype=np.ulonglong)
+    return pixel_math.empty_histogram(highest_healpix_order)
+
+
+def write_histogram(tmp_path, raw_histogram):
+    """overwrite existing raw histogram with updated values."""
+
+    file_name = os.path.join(tmp_path, "mapping_histogram.binary")
```
jeremykubica, line 27: There is an extra blank line here that can be removed.

jeremykubica (approved):

PR #32, opened 2023-02-23
Move file testing utilities to conftest.

delucchi-cmu:
Text files are loaded using a `with` statement.

Adds package extensions for pyarrow during linting. With this set of changes, there are no pylint findings for the `./tests/` tree (though I'm not adding to the CI pylint checks because I don't want to make everyone mad).


dougbrn (approved):
looks good to me

PR #37, opened 2023-03-07
Initial commit of import user guide.

delucchi-cmu:


OliviaLynn (approved):
Just want to check that [allwise_types.csv](https://github.com/astronomy-commons/hipscat-import/pull/37/files#diff-b425a4fb211889dc6c9624c258a99b420603dba1d0070aacdca5bcaea3aa0562) is intentionally left blank, even though the other two csv's are not (and the docs make them appear analogous). Otherwise, this looks good to me!

delucchi-cmu:
That was not intentional! Thanks for catching that!

PR #38, opened 2023-03-14
Unit test fixes

delucchi-cmu:
Recent changes to `hipscat` package cause tests to fail (and last night's smoke test caught it and emailed me! WAI!).

This addresses new epoch parameter and hive-style directory structures.

Additionally, introduces the pytest-timeout package. This will fail tests that take too long (if not marked for a longer timeout). This is a proxy for code performance monitoring (see https://github.com/lincc-frameworks/python-project-template/issues/79).


delucchi-cmu:
Also addresses a sneakier change that makes the `destination_pixel_map` contain integers.

Code coverage increased with recent commit. See https://app.codecov.io/github/astronomy-commons/hipscat-import/pull/38

tests/hipscat_import/test_map_reduce.py
```diff
@@ -201,7 +201,7 @@ def test_reduce_order0(parquet_shards_dir, assert_parquet_file_ids, tmp_path):
         id_column="id",
     )
 
-    output_file = os.path.join(tmp_path, "Norder0/Npix11", "catalog.parquet")
+    output_file = os.path.join(tmp_path, "Norder=0/Dir=0/Npix=11.parquet")
```
smcguire-cmu, line 204: I think when doing `os.path.join` each directory should be an argument to the method rather than having the slashes in the string. This is just for windows support though which we probably don't do anywhere else anyway.

tests/hipscat_import/test_run_import.py
```diff
@@ -112,7 +114,7 @@ def test_resume_dask_runner(
     assert_text_file_matches(expected_partition_lines, partition_filename)
 
     # Check that the catalog parquet file exists and contains correct object IDs
-    output_file = os.path.join(args.catalog_path, "Norder0/Npix11", "catalog.parquet")
+    output_file = os.path.join(args.catalog_path, "Norder=0", "Npix=11.parquet")
```
smcguire-cmu, line 117: Should there be a `Dir=` in here?

tests/hipscat_import/test_run_import.py
```diff
@@ -194,12 +198,13 @@ def test_dask_runner(
     assert_text_file_matches(expected_lines, metadata_filename)
 
     # Check that the catalog parquet file exists and contains correct object IDs
-    output_file = os.path.join(args.catalog_path, "Norder0/Npix11", "catalog.parquet")
+    output_file = os.path.join(args.catalog_path, "Norder=0/Dir=0/Npix=11.parquet")
```
smcguire-cmu, line 201: Same here with the `os.path.join`

tests/hipscat_import/test_run_import.py
```diff
@@ -251,6 +257,6 @@ def test_dask_runner_mixed_schema_csv(
     runner.run_with_client(args, dask_client)
 
     # Check that the catalog parquet file exists
-    output_file = os.path.join(args.catalog_path, "Norder0/Npix11", "catalog.parquet")
+    output_file = os.path.join(args.catalog_path, "Norder=0/Dir=0/Npix=11.parquet")
```
smcguire-cmu, line 260: And here

smcguire-cmu (approved):

PR #53, opened 2023-03-29
Enhancements to importing from FITS file.

delucchi-cmu:


src/hipscat_import/file_readers.py
```diff
@@ -122,14 +122,40 @@ def provenance_info(self) -> dict:
...
 
     def read(self, input_file):
-        """Read the whole fits file and return"""
-        yield Table.read(input_file, format="fits").to_pandas()
+        """Read chunks of rows in a fits file"""
+        table = Table.read(input_file, memmap=True)
+        if self.column_names:
+            table.keep_columns(self.column_names)
+        elif self.skip_column_names:
+            table.remove_columns(self.skip_column_names)
```
smcguire-cmu, line 143: Can you add a line in the docstring saying how this works? (i.e. how skip_column_names is ignored if both are defined)

src/hipscat_import/file_readers.py
```diff
@@ -122,14 +122,40 @@ def provenance_info(self) -> dict:
...
+
+        total_rows = len(table)
+        read_rows = 0
+
+        while read_rows < total_rows:
```
jeremykubica, line 148: What's the reason for breaking this up into chunks? Usually I see code do that to do some pruning on each chunk and keep the total used memory small. It looks like that is not the case here. So I'm guessing it is a limitation in to_pandas. Is that correct? If so, it might be worth adding a comment to that effect.

delucchi-cmu: The `memmap=True` in the table read avoids loading the whole file into memory at once. You're right that this is an attempt to keep the total memory down, since these FITS files are routinely >1G. Added a comment with link to more info.

Issue #42, opened 2023-03-14
Generate _hipscat_id field


delucchi-cmu:
The field we've been calling HOP (higher order pixel) or HOP+.

See https://docs.google.com/document/d/1v0U_5YB1WcpvnYB9Pg2TJ-0CWO2zJqAcGl8hoK1n1dc/edit

PR #56, opened 2023-03-30
Add hipscat index on import

delucchi-cmu:


delucchi-cmu:
Fixes #42

drewoldag (approved):
This looks ok to me. I don't have enough general knowledge to speak to the edge case coverage of the unit tests, so I have to trust you there :)

PR #59, opened 2023-03-30
Add console entry points for import tool.

delucchi-cmu:
With this, you can execute the import tool with commands like:

```
> hipscat-import --help
or
> hc --help
```


cchris28 (approved):
fine by me

Issue #47, opened 2023-03-14
Pipeline to create column-level indexes


delucchi-cmu:
This has some more requirements work to do to figure out if we want to include these in 1.0 (or at all)

PR #61, opened 2023-04-03
Move files for creation of new target catalogs.

delucchi-cmu:
This is a prefactor for issue #47 (create column-level indexes).

The new pipeline will reside in the import package, with its own map reduce and supporting files.


OliviaLynn (approved):
This looks good, always a fan of an elegant prefactor!

Issue #48, opened 2023-03-14
Source Tables - basic support


delucchi-cmu:
Support for writing source metadata.

PR #65, opened 2023-04-13
Updates to importer to support other table types.

delucchi-cmu:
Includes:

- catalog type in parameters
- test data for a source catalog
- fixes to map reduce to stop creating so many index columns
- adds columns to output so data is later readable easily as a dask dataframe (Norder, Dir, and Npix)
- write out the hipscat index field
- increases to test coverage

Closes Issue #48


OliviaLynn (approved):
Looks good!

PR #67, opened 2023-04-14
Prefactor for alternate table import arguments.

delucchi-cmu:
There are a handful of arguments that will be used for creating indexes and equijoin association tables. This pulls those arguments, their validation logic, and the provenance info generation into a parent class.

This should simplify the new classes, as well as reducing duplicated code (which pylint was VERY unhappy about).

### DIFF
--- a/src/hipscat_import/catalog/arguments.py
+++ b/src/hipscat_import/catalog/arguments.py
@@ -7,7 +7,8 @@ from typing import Callable, List
 
 import pandas as pd
 from hipscat.catalog import CatalogParameters
-from hipscat.io import file_io
+from hipscat.io import FilePointer, file_io
+from hipscat.pixel_math import hipscat_id
 
 from hipscat_import.catalog.file_readers import InputReader, get_file_reader
 from hipscat_import.runtime_arguments import RuntimeArguments
@@ -65,9 +66,10 @@ class ImportArguments(RuntimeArguments):
 
     epoch: str = "J2000"
     catalog_type: str = "object"
-    input_path: str = ""
+    input_path: FilePointer | None = None
     input_format: str = ""
-    input_file_list: List[str] = field(default_factory=list)
+    input_file_list: List[FilePointer] = field(default_factory=list)
+    input_paths: List[FilePointer] = field(default_factory=list)
 
     ra_column: str = "ra"
     dec_column: str = "dec"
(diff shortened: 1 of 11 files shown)


src/hipscat_import/catalog/arguments.py
```diff
@@ -58,156 +63,78 @@ class ImportArguments:
...
+    use_schema_file: str | None = None
+    resume: bool = False
+    highest_healpix_order: int = 10
+    pixel_threshold: int = 1_000_000
+    debug_stats_only: bool = False
+    filter_function: Callable | None = None
+    file_reader: InputReader | None = None
+
+    def __post_init__(self):
+        RuntimeArguments._check_arguments(self)
```
smcguire-cmu, line 85: Could this call happen in the `_check_arguments` method as a `super` call?

src/hipscat_import/catalog/arguments.py
```diff
@@ -58,156 +63,78 @@ class ImportArguments:
...
-        self.epoch = epoch
-        self.catalog_type = catalog_type
-        self._input_path = file_io.get_file_pointer_from_path(input_path)
-        self.input_format = input_format
-        self._input_file_list = (
-            [file_io.get_file_pointer_from_path(x) for x in input_file_list]
-            if input_file_list
+    epoch: str = "J2000"
+    catalog_type: str = "object"
+    input_path: str = ""
```
smcguire-cmu, line 68: Could this type be `FilePointer` if we're using it as one and it's being set to one in the post_init?

src/hipscat_import/catalog/arguments.py
```diff
@@ -58,156 +63,78 @@ class ImportArguments:
...
     def _check_arguments(self):
         """Check existence and consistency of argument values"""
-        if not self.catalog_name:
-            raise ValueError("catalog_name is required")
         if not self.input_format:
             raise ValueError("input_format is required")
-        if not self._output_path:
-            raise ValueError("output_path is required")
 
         if not 0 <= self.highest_healpix_order <= 19:
```
smcguire-cmu, line 104: Can we use this HiPSCat constant here instead of the literal?

src/hipscat_import/runtime_arguments.py
```diff
@@ -0,0 +1,115 @@
...
+        runtime_args.update(self.additional_provenance_info())
+        provenance_info = {
+            "tool_name": "hipscat_import",
+            "version": version("hipscat-import"),
+            "runtime_args": runtime_args,
+        }
+
+        return provenance_info
+
+    def additional_provenance_info(self):
```
smcguire-cmu, line 113: I think this method name is slightly confusing given that the additional info is added to the runtime args in the provenance info and not the provenance info itself

delucchi-cmu: Changed to `additional_runtime_provenance_info`. LMK if you think of a better name.

smcguire-cmu (approved):

PR #68, opened 2023-04-18
Write final parquet with pandas metadata.

delucchi-cmu:
Constructs the reduced table in pyarrow, and performs the final write with pandas, so the appropriate pandas metadata will be populated.


src/hipscat_import/catalog/map_reduce.py
```diff
@@ -212,7 +212,12 @@ def reduce_pixel_shards(
         "Npix",
         [np.full(rows_written, fill_value=destination_pixel_number, dtype=np.int32)],
     )
-    pq.write_table(merged_table, where=destination_file)
+    if add_hipscat_index:
+        merged_table.to_pandas().set_index("_hipscat_index").sort_index().to_parquet(
+            destination_file
+        )
+    else:
+        merged_table.to_pandas().to_parquet(destination_file)
```
smcguire-cmu, line 220: Does this not also save the pandas index column?

delucchi-cmu: It does now, if the pandas index looks like it's worth saving.

smcguire-cmu: Awesome, that seems like a good idea. Is that behavior documented anywhere? It'd be nice to warn users that would happen in a docstring/docs.

smcguire-cmu (approved):
Some documentation somewhere about how the pandas index removing works would be great, but other than that looks good!

PR #69, opened 2023-04-18
Initial implementation of association pipeline.

delucchi-cmu:


src/hipscat_import/association/map_reduce.py
```diff
@@ -0,0 +1,204 @@
...
+        table = pq.read_table(input_dir)
+        rows_written = len(table)
+
+        if rows_written != partition["num_rows"]:
+            raise ValueError(
+                "Unexpected number of objects ",
+                f" Expected {partition['num_rows']}, wrote {rows_written}",
+            )
+
+        pq.write_table(table, where=output_file)
```
smcguire-cmu, line 202: What is the table here going to be sorted by, if anything? I'm not sure if it really matters for performance, but I think writing with pandas metadata here makes sense, and the best index column is probably the primary hipscat index

src/hipscat_import/association/map_reduce.py
```diff
@@ -0,0 +1,204 @@
...
+            partition["Norder"],
+            partition["Npix"],
+            partition["join_Norder"],
+            partition["join_Npix"],
+        )
+        table = pq.read_table(input_dir)
+        rows_written = len(table)
+
+        if rows_written != partition["num_rows"]:
+            raise ValueError(
```
smcguire-cmu, line 197: Could we add a unit test for this case?

tests/hipscat_import/association/test_association_argument.py
```diff
@@ -0,0 +1,142 @@
...
+            primary_id_column="id",
+            primary_join_column="id",
+            join_input_catalog_path=small_sky_object_catalog,
+            join_id_column="id",
+            join_foreign_key="",  ## empty
+            output_path=tmp_path,
+            output_catalog_name="small_sky_self_join",
+            overwrite=True,
+        )
+
```
smcguire-cmu, line 96: Are the `output_path` and `output_catalog_name` not required?

src/hipscat_import/association/map_reduce.py
```diff
@@ -0,0 +1,204 @@
...
+        read_columns = [args.primary_join_column, args.primary_id_column] + read_columns
+
+    primary_index = dd.read_parquet(
+        path=args.primary_input_catalog_path,
+        columns=read_columns,
+        dataset={"partitioning": "hive"},
+    )
+    if single_primary_column:
+        ## Duplicate the column to simplify later steps
+        primary_index["primary_id"] = primary_index[args.primary_join_column]
```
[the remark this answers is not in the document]
delucchi-cmu: There are a handful of possibilities that could introduce weird behavior. The best thing might just be a warning that those column names are reserved for association tables.

src/hipscat_import/association/map_reduce.py
```diff
@@ -0,0 +1,204 @@
...
+        "Dir": "join_Dir",
+        "Npix": "join_Npix",
+    }
+    if not single_join_column:
+        rename_columns[args.join_id_column] = "join_id"
+    join_index = join_index.rename(columns=rename_columns).set_index("join_to_primary")
+
+    ## Join the two data sets on the shared join predicate.
+    join_data = primary_index.merge(
+        join_index, how="inner", left_index=True, right_index=True
```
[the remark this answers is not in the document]
delucchi-cmu: Doesn't hurt to add the param.

tests/hipscat_import/association/test_association_argument.py
delucchi-cmu: Yeah, they are, but I don't **have** to test it here to get coverage credit for it.

Thanks for keeping me honest.

tests/hipscat_import/association/test_association_argument.py
```diff
@@ -94,6 +94,109 @@ def test_empty_required(tmp_path, small_sky_object_catalog):
...
+            primary_id_column="id",
+            primary_join_column="id",
+            join_input_catalog_path=small_sky_object_catalog,
+            join_id_column="id",
+            join_foreign_key="id",
+            output_path=tmp_path,
+            output_catalog_name="bad_columns",
+            join_how="middle",  ## not a valid join option
+            overwrite=True,
+        )
```
smcguire-cmu, line 181: It'd be good if we could have a test for a valid alternative `join_how` to check that it works for other join methods

delucchi-cmu: Turns out, inner is really the only join type that makes sense for association tables, as written. Both sides should be present in order to build a valid directory path. Any other kinds of joins should be handled by downstream mechanisms.

src/hipscat_import/association/map_reduce.py
```diff
@@ -48,14 +47,15 @@ def map_association(args):
...
+    primary_index = (
+        primary_index.reset_index()
+        .rename(columns=rename_columns)
+        .set_index("primary_join")
     )
 
     ## Read and massage join input data
     single_join_column = args.join_id_column == args.join_foreign_key
     read_columns = [
-        "_hipscat_index",
```
smcguire-cmu: Is this because it's the index column?

smcguire-cmu (approved):

src/hipscat_import/association/map_reduce.py
delucchi-cmu: Yes - indexes aren't dataset columns.

PR #73, opened 2023-04-20
Add option to generate a catalog at a fixed order.

delucchi-cmu:


src/hipscat_import/catalog/arguments.py
```diff
@@ -77,8 +77,10 @@ class ImportArguments(RuntimeArguments):
     add_hipscat_index: bool = True
     use_schema_file: str | None = None
     resume: bool = False
+    constant_healpix_order: int = -1
```
drewoldag, line 80: Do these need to be added to the "attributes" section above?

src/hipscat_import/catalog/arguments.py
```diff
@@ -94,13 +96,30 @@ def _check_arguments(self):
...
-            )
-        if not 100 <= self.pixel_threshold <= 10_000_000:
-            raise ValueError("pixel_threshold should be between 100 and 10,000,000")
+        if self.constant_healpix_order >= 0:
+            if (
+                not 0
+                <= self.constant_healpix_order
+                <= hipscat_id.HIPSCAT_ID_HEALPIX_ORDER
+            ):
+                raise ValueError(
```
drewoldag, line 105: I see a couple of new tests, but this feels like a good amount of conditional logic. I'm assuming that there must be additional test cases that cover all of these.

I spent several minutes looking over this to try to identify a way to make this more concise, but didn't think of anything immediately that would help, given that the "between" logic is only used a few times, it's probably not work writing a separate function for it. 

I suppose that everyone has their preferred predicates, but how do you feel about using something like the following? 
```
if (
    self.constant_healpix_order < 0 or
    hipscat_id.HISPCAT_ID_HEALPIX_ORDER < self.constant_healpix_order
):
```

drewoldag (approved):
The logic looks fine. Left a couple small comments.

src/hipscat_import/catalog/arguments.py
delucchi-cmu: I have vague plans to add ANOTHER healpix_order parameter in the next ~week, and was going to put some shared range defaults/validation in at that point. But I'm happy to do it now instead!

There were no additional test cases to cover the logic, but I've added more tests for the helper method.

PR #76, opened 2023-04-27
Require dask_client if not requested.

delucchi-cmu:
When running a single test case, they can fail because no `dask_client` is instantiated, if there's no dask_client requested, but is implicity required for dask dataframe operations.

This requires a dask client fixture for all dask tests, even if `dask_client` is not among requested fixtures for the test case.

Also ups the default timeout, since 3 was pushing it for github workflows.


dougbrn (approved):
looks good to me!

PR #79, opened 2023-05-04
Improve performance of long-running imports by further splitting into map and split stages.

delucchi-cmu:
I ran against two sets of inputs on my local machine. Below are the runtimes, as well as counts of *intermediate* files:

```
                          old method      new method
1 input file: num files          497             255
              num dirs            24              15
              total time   0m36.147s       0m26.278s
4 input file: num files        50591            1424
              num dirs          2830              50
              total time   5m18.552s       1m53.075s
```

There is a large overhead to creating new intermediate parquet files, both in terms of time and disk space. For small sets, the improvement is minimal. As data sets grow, the improvement also grows.


maxwest-uw:
running on my `gaiaPartial` testing dataset (first 10 files of gaia), which contains ~2 million rows

`dask_n_workers` = 4

- **main** -> 9m 19.3s
- **mr_rehacktor** -> 15.7s\*
- **delucchi/splits** -> 41.3s

\*mr_rehacktor doesn't have reduction implemented, only the mapping phases.

a few things to note:
- the mapping in **delucchi/splits** takes pretty much the same time as my rehacktor
- the main difference is that the binning takes a bit longer in splits, although **mr_rehacktor** is somewhat stale and so there might be additional code functionality implemented there.

either way this is a great improvement and we're now within the same O(10 secs) runtime for my testing dataset 😎 I'll actually take a look at the code and see if there's any quick fixes to speed up anymore but so far this is awesome!!

delucchi-cmu:
More testing gave a warning about moving a large object (the alignment) onto the workers, and suggested using scatter. The latest commit uses that for the alignment object, as well as fails the whole pipeline when there are stage failures (which really should have been there the whole time and was making debugging frustrating).

maxwest-uw (approved):

PR #82, opened 2023-05-26
Max/margin generation

maxwest-uw:
The mapping step for margin cache generation. The reduce will be in a separate PR that comes after this one :)


src/hipscat_import/catalog/margin_cache_arguments.py
```diff
@@ -0,0 +1,94 @@
...
+        partition_stats = self.catalog.get_pixels()
+        highest_order = np.max(partition_stats["Norder"].values)
+        margin_pixel_k = highest_order + 1
+        if self.margin_order > -1:
+            if self.margin_order < margin_pixel_k:
+                # pylint: disable=line-too-long
+                raise ValueError(
+                    "margin_order must be of a higher order than the highest order catalog partition pixel."
+                )
+                # pylint: enable=line-too-long
```
delucchi-cmu, line 83:
```suggestion
                raise ValueError(
                    "margin_order must be of a higher order than "
                    "the highest order catalog partition pixel."
                )
```

And elsewhere. I don't know if there's an official recommendation, but I find the line breaks a lot easier to read, and results in fewer overall lines of code.

src/hipscat_import/catalog/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+            data[ra_column].values,
+            data[dec_column].values,
+            bounding_polygons
+        )
+
+    # pylint: disable-next=singleton-comparison
+    margin_data = data.loc[data["margin_check"] == True]
+
+    if len(margin_data):
+        # TODO: this should be a utility function in `hipscat`
```
delucchi-cmu, line 141: If this is just constructing a unique path for each shard, but the shards will be combined at a later stage and only the typical `/NOrder/Dir/NPix` will stick around, then keeping that logic local to the margin pipeline seems reasonable.

src/hipscat_import/catalog/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+                margin_threshold=args.margin_threshold,
+                output_path=args.margin_output_path,
+                margin_order=args.margin_order,
+                ra_column=args.catalog.catalog_info.ra_column,
+                dec_column=args.catalog.catalog_info.dec_column,
+            )
+        )
+
+    for _ in tqdm(
+        as_completed(futures, with_results=True),
```
delucchi-cmu, line 72: Will the mapping stage have a return value?

src/hipscat_import/catalog/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+        futures.append(
+            client.submit(
+                _map_pixel_shards,
+                partition_file=partition_file,
+                margin_pairs=margin_pairs,
+                margin_threshold=args.margin_threshold,
+                output_path=args.margin_output_path,
+                margin_order=args.margin_order,
+                ra_column=args.catalog.catalog_info.ra_column,
+                dec_column=args.catalog.catalog_info.dec_column,
```
delucchi-cmu, line 67: Noice.

src/hipscat_import/catalog/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+            partition_dir, source_order, source_pix
+        )
+
+        file_io.make_directory(shard_dir, exist_ok=True)
+
+        shard_path = paths.pixel_catalog_file(
+            partition_dir, source_order, source_pix
+        )
+
+        margin_data.to_parquet(shard_path)
```
delucchi-cmu, line 156: Will this parquet file get written out with all the extra columns added above? Will those get stripped out when we reduce the shards together, or could we just keep these notes in a separate data frame so it's clear that we're only writing out the original fields?

src/hipscat_import/catalog/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+import numpy as np
+import pandas as pd
+from dask.distributed import Client, as_completed
+from hipscat import pixel_math
+from hipscat.io import file_io, paths
+from tqdm import tqdm
+
+# pylint: disable=too-many-locals,too-many-arguments
+
+def _find_parition_margin_pixel_pairs(stats, margin_order):
```
delucchi-cmu, line 11:
```suggestion
def _find_partition_margin_pixel_pairs(stats, margin_order):
```

src/hipscat_import/catalog/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+
+def generate_margin_cache_with_client(client, args):
+    """Generate a margin cache for a given input catalog.
+    The input catalog must be in hipscat format.
+    Args:
+        client (dask.distributed.Client): A dask distributed client object.
+        args (MarginCacheArguments): A valid `MarginCacheArguments` object.
+    """
+    # determine which order to generate margin pixels for
+    partition_stats = args.catalog.get_pixels()
```
delucchi-cmu, line 227: Hmm.. Catalog should probably expose the healpixPixels list instead/as well, so users don't need to know the column names all the time.

https://github.com/astronomy-commons/hipscat/blob/main/src/hipscat/catalog/partition_info.py#L21

delucchi-cmu: Oh. You're using it below as `args.catalog.partition_info.get_healpix_pixels()`. Awesome. This should probably use that too for consistency, and so as not to use extra column name dependencies when you don't have to.

delucchi-cmu (commented):
So excited this is happening!

src/hipscat_import/catalog/__init__.py
```diff
@@ -2,26 +2,16 @@
...
-    set_reducing_done,
-    write_histogram,
-    write_mapping_done_key,
-    write_mapping_start_key,
-    write_reducing_key,
-)
+from .margin_cache import (generate_margin_cache,
+                           generate_margin_cache_with_client)
+from .margin_cache_arguments import MarginCacheArguments
+from .resume_files import (clean_resume_files, is_mapping_done,
```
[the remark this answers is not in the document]
maxwest-uw: it's the autoformatting from the pre-commit hooks 🤷‍♀️

src/hipscat_import/catalog/margin_cache.py
maxwest-uw: I don't have one set up for it atm

src/hipscat_import/catalog/margin_cache.py
maxwest-uw: good catch I'll remove them!

src/hipscat_import/catalog/margin_cache.py
maxwest-uw: is there something I should be returning? 😗

src/hipscat_import/catalog/margin_cache_arguments.py
```diff
@@ -0,0 +1,94 @@
...
+from hipscat.io import file_io, paths
+
+from hipscat_import.runtime_arguments import RuntimeArguments
+
+
+@dataclass
+class MarginCacheArguments(RuntimeArguments):
+    """Container for margin cache generation arguments
+    Attributes:
+        input_catalog_path (str): the path to the hipscat-formatted input catalog.
```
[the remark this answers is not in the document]
maxwest-uw: oh interesting! I kinda like that setup a lot more haha I'll go ahead and change it.

src/hipscat_import/margin_cache/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+            partition_dir, source_order, source_pix
+        )
+
+        file_io.make_directory(shard_dir, exist_ok=True)
+
+        shard_path = paths.pixel_catalog_file(
+            partition_dir, source_order, source_pix
+        )
+
+        final_df = margin_data.drop(columns=["partition_order", "partition_pixel", "margin_check"])
```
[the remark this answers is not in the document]
delucchi-cmu: Between the propagation of additional columns in the source data frame, and potential performance problems with exploding dataframes in this way, I'm wondering if it would be better to switch the logic to create a numpy mask, per margin pixel.

src/hipscat_import/margin_cache/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+        )
+    else:
+        data["margin_check"] = pixel_math.check_margin_bounds(
+            data[ra_column].values,
+            data[dec_column].values,
+            bounding_polygons
+        )
+
+    # pylint: disable-next=singleton-comparison
+    margin_data = data.loc[data["margin_check"] == True]
```
delucchi-cmu, line 138: margin_check doesn't get added in the case of polar. should it instead be

```
if is_polar:
  margin_data = _margin_stuff()
else:
  data["margin_check"] = ...
  margin_data = data...
```

src/hipscat_import/margin_cache/margin_cache.py
```diff
@@ -0,0 +1,249 @@
...
+        )
+
+    for _ in tqdm(
+        as_completed(futures, with_results=True),
+        desc="Mapping  ",
+        total=len(futures),
+    ):
+        ...
+
+def _map_pixel_shards(
```
delucchi-cmu, line 78: For organization, it would be nice to have `_map_pixel_shards` and everything it calls split out to another module. This one's already pretty big, and it would provide a clean unit for unit testing.

src/hipscat_import/margin_cache/margin_cache.py
maxwest-uw: this is how the code already works, `margin_check` will always be added, even in the polar case. see `_margin_filter_polar`

src/hipscat_import/margin_cache/margin_cache.py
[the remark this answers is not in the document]
maxwest-uw: `margin_check` is always added, good call on it not being removed. we can conditionally remove `is_trunc` as well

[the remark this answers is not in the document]
maxwest-uw: also I thought about the numpy mask idea a little more and it _might_ be possible to do but would be kind of rough and I'm not sure it would actual provide much use value for our case. we can't really mask on margin pixel because the pairings are many-to-many, so we could mask on which `margin_pixel`s belong to _a_ `partition_pixel` but not what pixel they should go to. I guess we could keep a mask for each partition pixel but that kinda makes our memory problem worse 😭

src/hipscat_import/catalog/__init__.py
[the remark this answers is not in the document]
maxwest-uw: `isort` specifically!

PR #91, opened 2023-06-22
Fleshing out guide for importing catalogs

delucchi-cmu:
In service of (but does not close) issue #35.

There are several TODOs here. I would like feedback on the overall structure before fiddling more with individual pages.


delucchi-cmu:
> not really super important but is there any benefit to using rst over markdown? I have a preference for markdown but it's also not a big deal

Pros of RST:
- we use several directives for cross-referencing pages that may not be well supported when using markdown under sphinx (e.g. table of contents, links between pages, links to auto-generated-apidocs, download links) (though they may be well-supported too, I just haven't tried them).
- we have to include another dependency in order to support mixing RST and MD files (https://github.com/astronomy-commons/hipscat/pull/99)
- because we can later add that dependency and mix RST and MD files in our docs, we are not limited to RST now or in the future

Though I certainly do appreciate the large Con that RST syntax is tedious and fiddly.

OliviaLynn (approved):
This looks good! 

I think there is something a little weird about having one of the suggested references in catalogs/ and the others in guides/ in index.rst:
```
:doc:`catalogs/arguments` (most common)
:doc:`guide/margin_cache`
:doc:`guide/association`
:doc:`guide/index_table`
```
but at the same time, it also makes sense to have the import arguments explanation live next to the catalogs one would import. So, no better method really jumps out at me.

PR #92, opened 2023-06-22
Add unit test for example in documentation.

delucchi-cmu:
## Change Description

In PR #91 , I claim that you can handle not-previously-seen-file-types with arguments like

```
        input_format="starr",
        file_reader=StarrReader(),
```

This is how I convinced myself that was true, and will prevent regression for documented behavior.


dougbrn (approved):
looks good to me!

PR #95, opened 2023-06-27
Pull resume logic into helper class.

delucchi-cmu:
## Solution Description

- Migrates handling of resume logging to a helper class
- Removes filter function 
- Scatters file reader on map/split methods


src/hipscat_import/catalog/resume_plan.py
```diff
@@ -0,0 +1,224 @@
...
+                if len(mapping_done_keys) != len(mapping_start_keys):
+                    raise ValueError(
+                        "Resume logs are corrupted. "
+                        "Delete temp directory and restart import pipeline."
+                    )
+                mapped_paths = set(mapping_start_keys)
+                self.map_files = [
+                    file_path
+                    for file_path in self.input_paths
+                    if f"map_{file_path}" not in mapped_paths
```
smcguire-cmu, line 98: Can generating the key like this be done in a separate function? Just to make sure it's consistent. (Not that it isn't right now, but in case we want to use it in the future)

src/hipscat_import/catalog/resume_plan.py
```diff
@@ -0,0 +1,224 @@
...
+        Otherwise, return histogram of appropriate shape."""
+        file_name = file_io.append_paths_to_pointer(
+            self.tmp_path, self.HISTOGRAM_BINARY_FILE
+        )
+        if file_io.does_file_or_directory_exist(file_name):
+            with open(file_name, "rb") as file_handle:
+                return frombuffer(file_handle.read(), dtype=np.int64)
+        return pixel_math.empty_histogram(highest_healpix_order)
+
+    def mark_mapping_done(self, mapping_key: str, histogram):
```
smcguire-cmu, line 126: I feel like this name is a little confusing given it's also writing the histogram as well as marking it done

src/hipscat_import/catalog/resume_plan.py
```diff
@@ -0,0 +1,224 @@
...
+        reduced_keys = set(self._read_log_keys(self.REDUCING_LOG_FILE))
+        reduce_items = [
+            (hp_pixel, source_pixels, f"{hp_pixel.order}_{hp_pixel.pixel}")
+            for hp_pixel, source_pixels in destination_pixel_map.items()
+        ]
+        reduce_items = [
+            (hp_pixel, source_pixels, key)
+            for (hp_pixel, source_pixels, key) in reduce_items
+            if key not in reduced_keys
+        ]
```
smcguire-cmu, line 174: This sets the same variable twice? I'm guessing the first time is the old version

smcguire-cmu (commented):
Looks good, I like the refactor! Few little things

delucchi-cmu (commented):
I had a child branch where I was doing even more refactoring for zubercal imports. I've merged that branch in, as it neatly addresses most of your comments.

src/hipscat_import/catalog/resume_plan.py
```diff
@@ -79,59 +80,84 @@ def gather_plan(self):
...
+
+        file_name = file_io.append_paths_to_pointer(
+            tmp_path, cls.HISTOGRAMS_DIR, f"{mapping_key}.binary"
+        )
         with open(file_name, "wb+") as file_handle:
             file_handle.write(histogram.data)
-        self._write_log_key(self.MAPPING_DONE_LOG_FILE, mapping_key)
+
+    def mark_mapping_done(self, mapping_key: str):
+        """Add mapping key to done list and update raw histogram"""
```
smcguire-cmu, line 159: I think this docstring needs to be updated with the change

src/hipscat_import/catalog/run_import.py
```diff
@@ -12,81 +12,67 @@
...
+        as_completed(futures),
         desc="Mapping  ",
         total=len(futures),
         disable=(not args.progress_bar),
     ):
         if future.status == "error":  # pragma: no cover
             some_error = True
-            continue
-        raw_histogram = np.add(raw_histogram, result)
-        resume.write_mapping_start_key(args.tmp_path, future.key)
```
smcguire-cmu, line 57: Where does the start_key get written now?

delucchi-cmu: There's no more distinction between start/done. The two keys were needed in case there was a failure within the histogram writing. In that case, we would have marked the shard as complete, but its raw histogram would not contribute to the overall histogram.

The shard's raw histogram is now written within the map method, so we don't need to worry about that edge case anymore.

src/hipscat_import/catalog/resume_plan.py
```diff
@@ -79,59 +80,84 @@ def gather_plan(self):
...
-                        "Resume logs are corrupted. "
-                        "Delete temp directory and restart import pipeline."
-                    )
-                mapped_paths = set(mapping_start_keys)
+                mapped_keys = set(self._read_log_keys(self.MAPPING_LOG_FILE))
                 self.map_files = [
-                    file_path
-                    for file_path in self.input_paths
-                    if f"map_{file_path}" not in mapped_paths
+                    (f"map_{i}", file_path)
```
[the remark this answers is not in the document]
delucchi-cmu: By making the keys shorter, we can use the key itself in sharded file names, and the keys need not be meaningful.

Added a step to make sure we're looking at the same files as the initial pipeline execution.

src/hipscat_import/catalog/run_import.py
smcguire-cmu: Oh cool, thanks!

smcguire-cmu (approved):

PR #96, opened 2023-06-28
Sam/manual dtypes

swyatt7:
## Change Description

Adding the ability to manually declare column dtypes for parquet metadata creation

## Solution Description

included argument `manual_dtypes` in the importer to accept a dictionary of `{"column_name":"dtype"}` that is explicitly applied onto the hipscat dataframes before writing. 

It is useful because declaring a schema doesn't prevent the parquet metadata from outputting an error like: 

```
RuntimeError: AppendRowGroups requires equal schemas.
```
When performing metadata collection. 

## Usage
argument is passed into `ImportArguments` before performing import runner
```
args = ImportArguments(
   ...
   manual_dtypes = {"column_name":"dtype"},
   ...
)
runner.run(args,...)
```


delucchi-cmu (commented):
Can you add some unit tests?

There's a similar test and some small test data already set up for the `use_schema_file` test case in `test_round_trip.test_import_mixed_schema_csv`.

delucchi-cmu:
Is this change still needed? We have alternative ways of dealing with unequal schema issues (https://hipscat-import.readthedocs.io/en/latest/notebooks/unequal_schema.html)

We should either clean this PR up and merge, or close it out.

PR #108, opened 2023-07-28
Initial implementation and tests for SOAP.

delucchi-cmu:
## Change Description

Creates a source-object-association-pipeline (SOAP) that is used to generate the join info between two catalogs with the specific object-source relationship.


src/hipscat_import/soap/map_reduce.py
```diff
@@ -0,0 +1,105 @@
...
+            cache_path, f"{source_healpix.order}_{source_healpix.pixel}.csv"
+        ),
+        index=False,
+    )
+
+
+def count_joins(soap_args, source_pixel, object_pixels, cache_path):
+    """Count the number of equijoined sources in the object pixels.
+    If any un-joined source pixels remain, stretch out to neighboring object pixels.
+    """
```
smcguire-cmu, line 47: Some type annotations or a description of args in the docstring would be useful here

src/hipscat_import/soap/map_reduce.py
```diff
@@ -0,0 +1,105 @@
...
+        if remaining_sources < 1:
+            break
+        join_count = _count_joins_for_object(
+            source_data,
+            soap_args.object_catalog_dir,
+            soap_args.object_id_column,
+            object_pixel,
+        )
+        results.append([object_pixel.order, object_pixel.pixel, join_count])
+        remaining_sources -= join_count
```
smcguire-cmu, line 70: If the join_count is 0 and there are no matches between the pixels, will it still get added to the file?

src/hipscat_import/soap/resume_plan.py
```diff
@@ -0,0 +1,155 @@
...
+    """Build a map of (source order/pixel) to the (object order/pixel)
+    that are aligned, as well as neighboring object pixels.
+    """
+    object_catalog = Catalog.read_from_hipscat(args.object_catalog_dir)
+    source_catalog = Catalog.read_from_hipscat(args.source_catalog_dir)
+
+    ## Direct aligment from source to object
+    ###############################################
+    grouped_alignment = align_trees(
+        object_catalog.pixel_tree, source_catalog.pixel_tree, "outer"
```
smcguire-cmu, line 94: I know it probably shouldn't happen for a source and object catalog, but since it's an outer alignment what happens  if one of the pixels is None?

src/hipscat_import/soap/resume_plan.py
```diff
@@ -0,0 +1,155 @@
...
+            HealpixPixel(int(source_name[0]), int(source_name[1])),
+            [
+                HealpixPixel(int(object_elem[0]), int(object_elem[1]))
+                for object_elem in object_group.dropna().to_numpy().T[:2].T
+            ],
+        )
+        for source_name, object_group in grouped_alignment
+    ]
+    ## Treat the array of tuples as a dictionary.
+    source_to_object = dict(source_to_object)
```
smcguire-cmu, line 116: I did not know you could do this to make a dict, neat

src/hipscat_import/soap/resume_plan.py
```diff
@@ -0,0 +1,155 @@
...
+        object_order_map[exploded_pixels] = pixel.order
+
+    for source, objects in source_to_object.items():
+        # get all neighboring pixels
+        nside = hp.order2nside(source.order)
+        neighbors = hp.get_all_neighbours(nside, source.pixel, nest=True)
+
+        ## get rid of -1s and normalize to max order
+        explosion_factor = 4 ** (max_order - source.order)
+        neighbors = [neighbor * explosion_factor for neighbor in neighbors if neighbor != -1]
```
smcguire-cmu, line 144: I think this isn't exactly getting the neighboring pixels after it's exploded, since there are `explosion_factor` number of pixels at a higher order, and this just computes the first of those, and the first pixel is always the top one. 

I don't think this matters after it has been desploded unless the object order is higher than the source order which probably shouldn't happen

smcguire-cmu: Having said that, I think Max has code in hipscat that gets the correct neighboring pixels as part of the margin cache

src/hipscat_import/soap/resume_plan.py
```diff
@@ -0,0 +1,155 @@
...
+
+        neighbors_orders = object_order_map[neighbors]
+        desploded = [
```
smcguire-cmu, line 147: 2 things:

1. Desploded is a great name
2. I like the map at a higher order to get the matching object pixel, that's smart :)

src/hipscat_import/soap/map_reduce.py
```diff
@@ -0,0 +1,105 @@
...
+from hipscat.io.paths import pixel_catalog_file
+
+
+def _count_joins_for_object(source_data, object_catalog_dir, object_id_column, object_pixel):
+    object_path = pixel_catalog_file(
+        catalog_base_dir=object_catalog_dir,
+        pixel_order=object_pixel.order,
+        pixel_number=object_pixel.pixel,
+    )
+    object_data = pd.read_parquet(path=object_path, columns=[object_id_column]).set_index(object_id_column)
```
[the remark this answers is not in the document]
smcguire-cmu: If we have, this could do with a wrapper method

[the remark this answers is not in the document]
smcguire-cmu: Max did just merge a method for this if you don't mind replacing the read_parquet? https://github.com/astronomy-commons/hipscat/pull/123

src/hipscat_import/soap/map_reduce.py
smcguire-cmu: Is this desired behavior? I would have thought that if there are no joins for an object pixel we wouldn't want to include it

smcguire-cmu (commented):
I still have a few things from my last review I'm uncertain of, would you mind explaining your thoughts on them?

src/hipscat_import/soap/map_reduce.py
delucchi-cmu: It won't be included in the final output.

src/hipscat_import/soap/map_reduce.py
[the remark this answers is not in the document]
smcguire-cmu: Max actually made that method in a PR a few days ago. https://github.com/astronomy-commons/hipscat/pull/123

But it can wait for another PR to replace all the read_parquet

src/hipscat_import/soap/map_reduce.py
smcguire-cmu: Huh, your previous comment from 3 days ago didn't show up for me when I reviewed this yesterday. Sorry for unresolving this, it makes sense.

smcguire-cmu (approved):
Looks good! For some reason your previous comments didn't show up for me, so sorry about asking for more clarification

PR #109, opened 2023-07-28
Consolidate logic to wait for dask futures.

delucchi-cmu:
## Change Description

Reducing code duplication for waiting on dask futures, updating a tqdm task bar, dask task log, raises error, and marks stage done. Also switches from explicit file names to single stage name.

Code duplication was raised as a pylint error in the SOAP PR, so I'm trying to get ahead of it.

### DIFF
--- a/src/hipscat_import/catalog/run_import.py
+++ b/src/hipscat_import/catalog/run_import.py
@@ -7,7 +7,6 @@ The actual logic of the map reduce is in the `map_reduce.py` file.
 
 import hipscat.io.write_metadata as io
 import numpy as np
-from dask.distributed import as_completed
 from hipscat import pixel_math
 from tqdm import tqdm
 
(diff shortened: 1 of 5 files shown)


src/hipscat_import/catalog/run_import.py
drewoldag, line 1: I like the reduction of code here :)

drewoldag (approved):
This looks pretty cool. My only question here is the same as the one I posted in slack - I'm curious at what point you would make the decision to move away from `dataclass` to a traditional class this work.

delucchi-cmu:
I hadn't been thinking too much about it, but now that you're asking me, I guess I will:

I'm in favor of keeping things plain-old-dataclasses for now, and for the foreseeable future. In nearly all cases, the methods that are being added are not modifying the state of the member variables, but are rather helper methods that could live just about anywhere. For the sake of code organization, it makes sense for them to live as class methods.

And I think to your question on slack

    does the reduction of boilerplate that dataclass provides imply that maybe more of our traditional classes should be dataclasses

I think that's probably true, though there are certainly limitations and many things that will need to stay full-fledged classes. I have some biases that are showing (coming from years in java persistence layers, and years with protocol buffers) that make me lean toward these kind of container objects.

PR #117, opened 2023-08-11
Refactor based on hipscat v0.1.2

delucchi-cmu:
## Change Description

Uses the new `file_io.load_parquet_to_pandas` method, and test-friendly pixel list construction of a `Catalog`.

### DIFF
--- a/src/hipscat_import/catalog/file_readers.py
+++ b/src/hipscat_import/catalog/file_readers.py
@@ -2,11 +2,10 @@
 
 import abc
 
-import pandas as pd
 import pyarrow as pa
 import pyarrow.parquet as pq
 from astropy.table import Table
-from hipscat.io import file_io
+from hipscat.io import FilePointer, file_io
 
 # pylint: disable=too-few-public-methods,too-many-arguments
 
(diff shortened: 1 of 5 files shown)


src/hipscat_import/catalog/file_readers.py
```diff
@@ -141,7 +140,9 @@ def read(self, input_file):
         self.regular_file_exists(input_file)
 
         if self.schema_file:
-            schema_parquet = pd.read_parquet(self.schema_file, dtype_backend="numpy_nullable")
+            schema_parquet = file_io.load_parquet_to_pandas(
+                FilePointer(self.schema_file), dtype_backend="numpy_nullable"
```
maxwest-uw, line 144: do the api changes mean that we have to cast any path strings to `FilePointer`s for hipscat functions to work? not that it's bad to be explicit.

maxwest-uw (approved):
had one small comment about something that falls outside of the scope of this pr :) lgtm

src/hipscat_import/catalog/file_readers.py
delucchi-cmu: Functionally, I think the methods might work with a string argument? But because this is typed code, we'll fail type checks.

PR #143, opened 2023-10-06
Resume file prefactor

delucchi-cmu:
This prefactor re-arranges some expectations around resume files.

- each task should now mark itself done before returning from the task method
- no more log files
- the done state should be marked in a file that matches the task's key. this is either by "done" files, or another small file of intermediate results (we already had some of these files, so they're now doing double-duty)
- makes a distinction between the resume_path where these small resume files go, and cache_shard_path, where sharded parquet files go. they're the same path right now, but i'd like to change that in the future.
- leaves in the waiting_for_futures kind of functionality, just so we still get a progress bar

## Code Quality


src/hipscat_import/catalog/resume_plan.py
```diff
@@ -129,15 +170,20 @@ def read_histogram(self, healpix_order):
     @classmethod
     def write_partial_histogram(cls, tmp_path, mapping_key: str, histogram):
         """Write partial histogram to a special intermediate directory"""
-        file_io.make_directory(
```
camposandro, line 132: [`map_to_pixels`](https://github.com/astronomy-commons/hipscat-import/blob/5ddb60c598028279f490a66184becc111d3b234f/src/hipscat_import/catalog/map_reduce.py#L114C8-L114C8) calls this method but it does not verify that the histogram's directory exists. Should we keep this make_directory call to make sure no error is thrown?

camposandro (commented):
It was awesome to review this PR - great refactor!

camposandro (approved):

PR #167, opened 2023-11-20
Add tests for more complex sorting scenarios.

delucchi-cmu:
## Change Description

This doesn't make any functional changes to the catalog import process. This just provides demonstration (proof?) of discussions around tweaking sort options during catalog import to achieve particular ordering of resulting parquet files.

### DIFF
--- a/tests/.pylintrc
+++ b/tests/.pylintrc
@@ -333,7 +333,7 @@ indent-string='    '
 max-line-length=110
 
 # Maximum number of lines in a module.
-max-module-lines=500
+max-module-lines=1000
 
 # Allow the body of a class to be on the same line as the declaration if body
 # contains single statement.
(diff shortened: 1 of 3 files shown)


hombit (approved):
Thank you!

It would be helpful to check that sources are still sorted by `_hipscat_index`.

hombit:
The other (or additional) way to test it is doing something like the following:
1) Concatenate pandas data frames,
2) sort them by (ra, dec, source_id) in the first case and by (ra, dec, object_id, time) in the second case,
3) and compare hipcat files with these sorted dataframes (skipping hipscat-specific columns like `_hipscat_index`)

delucchi-cmu:
> The other (or additional) way to test it is doing something like the following:
> 
> 1. Concatenate pandas data frames,
> 2. sort them by (healpy.ang2pix(ra, dec), source_id) in the first case and by (healpy.ang2pix(ra, dec), object_id, time) in the second case,
> 3. and compare hipcat files with these sorted dataframes (skipping hipscat-specific columns like `_hipscat_index`)

Added dataframe comparisons alongside more boring series comparisons.

PR #168, opened 2023-11-21
Argument renaming

delucchi-cmu:
## Change Description

Four refactors bundled up here:
- renames the argument `output_catalog_name` to `output_artifact_name` (see https://github.com/astronomy-commons/hipscat-import/pull/126)
- renames catalog import argument `id_column` to `sort_columns` (this field need not be an id (could be time), and can support more than one column)
- clarifies the type for `sort_columns` (was mixing `str` and `List[str]`)
- runs black and isort

### DIFF
--- a/src/hipscat_import/catalog/map_reduce.py
+++ b/src/hipscat_import/catalog/map_reduce.py
@@ -79,7 +79,7 @@ def _iterate_input_file(
                 )
             # Set up the pixel data
             mapped_pixels = hp.ang2pix(
-                2**highest_order,
+                2 ** highest_order,
                 data[ra_column].values,
                 data[dec_column].values,
                 lonlat=True,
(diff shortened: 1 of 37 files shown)


src/hipscat_import/catalog/map_reduce.py
```diff
@@ -273,8 +273,8 @@ def reduce_pixel_shards(
         )
 
     dataframe = merged_table.to_pandas()
-    if id_column:
-        dataframe = dataframe.sort_values(id_column)
+    if sort_columns:
+        dataframe = dataframe.sort_values(sort_columns.split(","))
```
delucchi-cmu, line 277: This is the sole behavioral change in the PR.

hombit (approved):
I don't understand the context of catalog -> artifact, but everything looks great to me!

Issue #172, opened 2023-11-28
Doc build failing


delucchi-cmu:
The documentation build is failing on in-progress PRs, but for reasons that appear unrelated to the PRs.

PR #173, opened 2023-11-28
Give more failure info in waiting for futures

delucchi-cmu:
Might give more context in debugging issue #172

### DIFF
--- a/src/hipscat_import/pipeline.py
+++ b/src/hipscat_import/pipeline.py
@@ -63,7 +63,7 @@ def pipeline_with_client(args: RuntimeArguments, client: Client):
 
 def _send_failure_email(args: RuntimeArguments, exception: Exception):
     if not args.completion_email_address:
-        raise exception
+        raise exception from None
     message = EmailMessage()
     message["Subject"] = "hipscat-import failure."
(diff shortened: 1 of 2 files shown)


src/hipscat_import/pipeline.py
```diff
@@ -63,7 +63,7 @@ def pipeline_with_client(args: RuntimeArguments, client: Client):
 
 def _send_failure_email(args: RuntimeArguments, exception: Exception):
     if not args.completion_email_address:
-        raise exception
+        raise exception from None
```
hombit, line 66: I'm a bit confused here. Why would we like to raise the same exception but unchain it? Should we raise another exception about the lack of `args.completion_email_address`?

hombit (approved):

src/hipscat_import/pipeline.py
delucchi-cmu: Fair. There's an additional method call in the stack trace because the exception isn't re-raised until the `_send_failure_email` method. I thought about adding a few more comments in the code to make that clearer, but instead re-structured the exception handling to hopefully make the intent clearer and to avoid the additional method in the stack trace.

PR #174, opened 2023-11-29
Remove the explicit task key.

### LINKED ISSUE #172: Doc build failing

delucchi-cmu:
## Change Description

Fixes #172 .

## Solution Description

My theory on what's going on here is that dask sees that it already has some completed tasks with the name `split_<k>`, and so doesn't even try to run them when they're encountered in the second pipeline execution. PR #173 tried to work around this issue by re-initializing the client to clear out the tasks, but that doesn't feel good/sustainable.

We *were* using the future's key to log completed stages after task success, but PR #143 took that out, so we can just let dask assign its own task keys.

### DIFF
--- a/docs/notebooks/unequal_schema.ipynb
+++ b/docs/notebooks/unequal_schema.ipynb
@@ -82,12 +82,12 @@
     "    highest_healpix_order=1,\n",
     ")\n",
     "\n",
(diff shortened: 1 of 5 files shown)


dougbrn (approved):
LGTM

PR #181, opened 2023-12-04
Optionally write out all of the parquet files.

delucchi-cmu:
## Change Description

This is part of https://github.com/astronomy-commons/hipscat/issues/147 (in a kind of round-a-bout way).

This adds functionality to the pipeline to optionally write out all of the leaf parquet files necessary for the association table, mostly to generate a test association catalog for hipscat/lsdb. When leaf parquet files are written out, we use those footers to generate a `_metadata` file for the catalog. Otherwise, we artificially create the necessary metadata.


src/hipscat_import/soap/run_soap.py
```diff
@@ -22,28 +22,50 @@ def run(args, client):
...
             futures.append(
                 client.submit(
                     count_joins,
+                    key=source_key,
                     soap_args=args,
                     source_pixel=source_pixel,
                     object_pixels=object_pixels,
-                    cache_path=args.tmp_path,
                 )
             )
```
camposandro, line 34: hey @delucchi-cmu, last week you removed the task’s key here because it was causing issues with the full pipeline execution (#174). Is it supposed to be back, or was it readded by mistake?

src/hipscat_import/soap/run_soap.py
```diff
@@ -22,28 +22,50 @@ def run(args, client):
...
 
         resume_plan.wait_for_counting(futures)
 
+    if args.write_leaf_files and not resume_plan.is_reducing_done():
+        for object_pixel, object_key in resume_plan.reduce_keys:
+            futures.append(
+                client.submit(
+                    reduce_joins,
+                    key=object_key,
```
camposandro, line 43: Would this key also pose a problem?

camposandro (commented):
Other than that, it all seems to make sense to me!

src/hipscat_import/soap/run_soap.py
delucchi-cmu: Great catch! That was re-added by copy pasta mistake.

src/hipscat_import/soap/run_soap.py
delucchi-cmu: Same mistake here.

camposandro (approved):

PR #202, opened 2024-01-19
Performance improvements for index generation

delucchi-cmu:
## Change Description

Related to https://github.com/astronomy-commons/lsdb/issues/108

This improves the performance of generating indexes for very large datasets.

- adds optional division hints for setting the new index
- adds option to skip `drop_duplicates` operation
- sneaks in a quality-of-life improvement when running `pytest` on epyc: the last test would often fail as a timeout.


src/hipscat_import/index/map_reduce.py
```diff
@@ -31,16 +31,38 @@ def create_index(args):
...
-    data = data.set_index(args.indexing_column)
+
+    if args.divisions is not None and len(args.divisions) > 2:
+        data = data.set_index(args.indexing_column, divisions=args.divisions)
+    else:
+        data = data.set_index(args.indexing_column)
+
+    if args.drop_duplicates:
+        data = (
+            data.repartition(partition_size=1_000_000_000)
```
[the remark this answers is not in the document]
delucchi-cmu: This is a weird one. And I was half-hoping no one would notice so I wouldn't have to try to explain what's happening =D

Dask uses pyarrow for the drop_duplicates step. Pyarrow has pretty strict limit on the size that its arrays can grow to. If *any* of the columns contain strings, then pyarrow's memory just goes through the roof. So this re-partitioning in particular is just for the pyarrow-specific memory limit (that doesn't seem to have any relation to the total memory of the machine you're working on). 

Once the de-duplication step is complete, we use the user-provided partition size for computing the on-disk partitions, and that can be larger again.

Or if we don't need to do any de-duplication, then we don't need to worry about the pyarrow memory limit issue at all. At least this is what I've observed in the handful of cases I've tried.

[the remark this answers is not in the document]
camposandro: Ah, understood! Out of curiosity, is this 1GiB memory limit specified anywhere in their documentation or did you come up with it after running different configurations? I couldn't find any reference for it.

src/hipscat_import/index/map_reduce.py
```diff
@@ -31,16 +31,38 @@ def create_index(args):
...
+            .reset_index()
+            .drop_duplicates()
+            .set_index(args.indexing_column, sorted=True)
```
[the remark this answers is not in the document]
delucchi-cmu: You're wise beyond your years.

src/hipscat_import/index/map_reduce.py
```diff
@@ -31,16 +31,38 @@ def create_index(args):
...
     if not args.include_hipscat_index:
         data = data.drop(columns=[HIPSCAT_ID_COLUMN])
-    data = data.drop_duplicates()
-    data = data.repartition(partition_size=args.compute_partition_size)
-    data = data.set_index(args.indexing_column)
+
+    if args.divisions is not None and len(args.divisions) > 2:
+        data = data.set_index(args.indexing_column, divisions=args.divisions)
+    else:
+        data = data.set_index(args.indexing_column)
```
dougbrn, line 55: Just in case you aren't aware, with default parameters this set_index call will be sorting by the new index column as well.  https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.set_index.html

delucchi-cmu: Yes! It just won't have the divisions to use as hints, and so will have much worse performance for large data sets.

src/hipscat_import/index/map_reduce.py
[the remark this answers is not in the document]
delucchi-cmu: Added more comments.

src/hipscat_import/index/map_reduce.py
```diff
@@ -31,16 +31,44 @@ def create_index(args):
...
+        #   pyarrow in the drop_duplicates implementation (
+        #   "array cannot contain more than 2147483646 bytes")
+        # - Reset the index, so the indexing_column values can be considered
+        #   when de-duping.
+        # - Drop duplicate rows
+        # - Set the index back to our indexing_column, but this time, the
+        #   values are still sorted so it's cheaper.
+        data = (
+            data.repartition(partition_size=1_000_000_000)
+            .reset_index()
```
dougbrn, line 67: These reset_index->set_index operations seem expensive generally, though it's nice you're able to set the sorted flag. It seems like drop_duplicates doesn't have a way to consider the index, which makes sense why we have to do this. I wonder if there's some way to use groupby to get the duplicated indices, and apply a function that dedupes from each group. It may be a toss-up whether having a groupby or a set_index is more performance intensive though...

src/hipscat_import/index/map_reduce.py
```diff
@@ -31,16 +31,46 @@ def create_index(args):
...
+            .drop_duplicates()
+            .set_index(args.indexing_column, sorted=True)
+        )
+
+    # Now just write it out to leaf parquet files!
     data = data.repartition(partition_size=args.compute_partition_size)
```
smcguire-cmu, line 73: I wonder if we can get away without doing a repartition after the `set_index`? That the `set_index` method has the `npartitions` and `partition_size` parameters. Maybe doing the repartitioning with those as part of the set index call could be quicker than having to do two passes over the data for repartitioning? It could just be doing a separate repartition call under the hood with those parameters in `set_index` so it might not make a difference.

smcguire-cmu: Also is it ok that here if the user specifies `divisions` in the args and it's used to set the index, with this repartition I think it will ignore those divisions and just do the divisions to the partition_size? I don't think that's necessarily a problem, but as a user it might be confusing to set the divisions and then have the output files not follow those divisions.

src/hipscat_import/index/map_reduce.py
smcguire-cmu: This might be very wrong, I think I have some questions about duplicates

delucchi-cmu: After spending an hour on this, I've decided it's a dask rabbit-hole. Group by and other aggregators are subtly different between pandas and the dask dataframe implementation (often with totally good reason). If it's alright, I'm going to open an issue to investigate further (that someone else is welcome to take on!) so we can move forward with the other incremental improvements from this PR.

https://github.com/astronomy-commons/hipscat-import/issues/206

dougbrn (approved):
Thanks for spending some time on the groupby suggestion, I think this looks good!

PR #205, opened 2024-01-30
Fix margin cache generation leaving empty shard dirs

smcguire-cmu:
## Change Description
Fixes a bug where generating a margin cache left the empty top level shard directories.

## Solution Description
In a previous PR, the margin cache shards were changed to be generated from the final pixel destination `Norder=x/Dir=y/Npix=z/...` to the same shard directories as other pipelines `order_x/dir_y/pix_z/...`. Before, these directories did not need to be deleted since they were used as the final structure, but now the new shard directory structures do need to be deleted.

To solve this, the shard files are now stored under a subdirectory `cache/order_x/...` which is deleted at the end of the pipeline.

## Code Quality

### DIFF
--- a/src/hipscat_import/margin_cache/margin_cache.py
+++ b/src/hipscat_import/margin_cache/margin_cache.py
@@ -145,5 +145,7 @@ def generate_margin_cache(args, client):
             catalog_base_dir=args.catalog_path, dataset_info=margin_catalog_info
         )
         step_progress.update(1)
+        cache_path = mcmr.get_cache_directory(args.catalog_path)
         file_io.remove_directory(args.tmp_path, ignore_errors=True)
+        file_io.remove_directory(cache_path, ignore_errors=True)
         step_progress.update(1)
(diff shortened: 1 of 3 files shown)


src/hipscat_import/margin_cache/margin_cache.py
```diff
@@ -145,5 +145,7 @@ def generate_margin_cache(args, client):
             catalog_base_dir=args.catalog_path, dataset_info=margin_catalog_info
         )
         step_progress.update(1)
+        cache_path = mcmr.get_cache_directory(args.catalog_path)
         file_io.remove_directory(args.tmp_path, ignore_errors=True)
+        file_io.remove_directory(cache_path, ignore_errors=True)
```
delucchi-cmu, line 150: Hm. It looks like the margin cache process doesn't make use of the `args.tmp_path` at all, and would have been cleaned up by the above step, if it were. I created a branch that uses the `tmp_path` instead (https://github.com/astronomy-commons/hipscat-import/tree/delucchi/margin_tmp). Can you see if that addresses the empty shard dir problem? I'd like users to be able to use a different path for tmp files (see this doc: https://hipscat-import.readthedocs.io/en/latest/catalogs/temp_files.html)

smcguire-cmu: Ah I see! That makes sense, and yeah I tested with that branch and it solves the issue. I can close this PR and we can go with that solution if you think that's better.

delucchi-cmu: Sounds good. Sending you the PR.

smcguire-cmu:
#207 fixes the same issue as this and uses the tmp_path for intermediate files, so no need for this.

PR #212, opened 2024-02-09
Expand docs for margin and index generation.

delucchi-cmu:
## Change Description

Gives all the help I can think to give when generating margin caches or index tables.

Note that this duplicates A LOT of the text between `catalogs/arguments.rst`, `guide/margin_cache.rst` and `guide/index_table.rst`. This is intentional, and based on feedback I got from Troy about duplicating some argument specifications instead of making users look in a few different places to put their pipeline together.

### DIFF
--- a/docs/catalogs/arguments.rst
+++ b/docs/catalogs/arguments.rst
@@ -220,6 +220,7 @@ There are several stages to the pipeline execution, and you can expect progress
 reporting to look like the following:
 
 .. code-block::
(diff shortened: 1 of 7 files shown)


delucchi-cmu:
Docs previews:

https://hipscat-import--212.org.readthedocs.build/en/212/guide/margin_cache.html
https://hipscat-import--212.org.readthedocs.build/en/212/guide/index_table.html

PR #222, opened 2024-02-15
Unit test and schema propagation for gaia import.

delucchi-cmu:
## Change Description

I've encountered some resistance to importing a subset of gaia data on epyc.

The CSV files begin with rows like 
```
# blah blah ignore these lines
# no really, they aren't relevant
id, ra, dec
45, 5,6
...
```

We should be able to add the single `comment="#"` kwarg to the CsvReader, and everything is ok.

But. The small subset I'm trying to import ends up with several small partitions that have final parquet files where the a particular column has only empty values, and should instead be a string. This gets the annoying pyarrow unequal schemas error. The CsvReader couldn't handle having BOTH a provided schema AND some kwargs.

Also, the schema would be used when re-reading the parquet shards back in, but if the final, combined parquet file STILL had empty columns, pandas would ignore the schema. So this adds logic to add the schema to the writing call.

BUT. If the schema that we provide already has a `__index_level_0__` pandas index field, then this *might* not match the expected type of the `_hipscat_index`, and the `_hipscat_index` really does require all 64 bits of the uint64.


src/hipscat_import/catalog/file_readers.py
```diff
@@ -126,6 +126,7 @@ def __init__(
         column_names=None,
         type_map=None,
         separator=",",
+        parquet_kwargs=None,
```
hombit, line 129: Could you please add "Parameters" to the class-level docstring to explain the difference between `kwargs` and `parquet_kwargs`

hombit (approved):
I cannot say I fully understand the PR, but everything looks reasonable, and the imported files look fine

src/hipscat_import/catalog/file_readers.py
hombit: Thank you for fixing it!

tests/hipscat_import/catalog/test_run_round_trip.py
```diff
@@ -411,3 +411,51 @@ def test_import_hipscat_index(
...
+    input_file = os.path.join(formats_dir, "gaia_minimum.csv")
+    schema_file = os.path.join(formats_dir, "gaia_minimum_schema.parquet")
+
+    args = ImportArguments(
+        output_artifact_name="gaia_minimum",
+        input_file_list=[input_file],
+        input_format="csv.gz",
+        file_reader=CsvReader(
+            comment="#",
+            schema_file=schema_file,
```
hombit, line 432: I've got an error when I was trying to run the same code with file `GaiaSource_310878-313367.csv.gz`. However adding `parquet_kwargs={"dtype_backend": "numpy_nullable"}` helped me to read the file

hombit:
I think I should write it here instead of creating a new issue, it is the only blocker right now
https://github.com/astronomy-commons/hipscat-import/issues/227

PR #232, opened 2024-02-21
Use primary catalog schema for margin cache.

### LINKED ISSUE #230: Margin generation should use primary catalog's schema
**Bug report**

Sparse data in a sparse table like margin catalogs can lead to mismatched schema issues. Additionally, there may be key-value metadata in the primary catalog's schema that should be carried through to the margin catalog.

e.g. this error when creating the margin cache for `allwise`:

```
AppendRowGroups requires equal schemas.
The two columns with index 79 differ.
column descriptor = {
  name: w3cc_map_str,
  path: w3cc_map_str,
  physical_type: INT32,
  converted_type: NONE,
  logical_type: Null,
  max_definition_level: 1,
  max_repetition_level: 0,
}
column descriptor = {
  name: w3cc_map_str,
  path: w3cc_map_str,
  physical_type: BYTE_ARRAY,
  converted_type: UTF8,
  logical_type: String,
  max_definition_level: 1,
  max_repetition_level: 0,
}
```

Pretty sure we can use the `<primary catalog>/_common_metadata` file for this purpose.

**Before submitting**
Please check the following:

delucchi-cmu:
## Change Description

Closes #230


src/hipscat_import/margin_cache/margin_cache_map_reduce.py
```diff
@@ -90,13 +92,16 @@ def _to_pixel_shard(data, margin_threshold, output_path, ra_column, dec_column):
                 PartitionInfo.METADATA_PIXEL_COLUMN_NAME: np.uint64,
             }
         )
+        final_df = final_df.set_index(HIPSCAT_ID_COLUMN).sort_index()
```
smcguire-cmu, line 95: This means the margin cache now has the hipscat_index, right?

smcguire-cmu (approved):
Looks good!

src/hipscat_import/margin_cache/margin_cache_map_reduce.py
delucchi-cmu: Yes - future margin caches will carry through the `_hipscat_index` of their primary catalog.

PR #234, opened 2024-02-21
Remove input_format argument.

### LINKED ISSUE #215: Correct tutorial on when input_format argument is required
The [Reading input files tutorial](https://hipscat-import.readthedocs.io/en/latest/catalogs/arguments.html#which-files) seems to suggest that the `input_format` argument is only needed when the user provides `input_path` rather than `input_file_list`.

However hipscat-import seems to always fail argument validation when `input_format` is not provided https://github.com/astronomy-commons/hipscat-import/blob/6eb5204166e942902de81625062e3ed91e84c93f/src/hipscat_import/catalog/arguments.py#L91

### LINKED ISSUE #229: Format - file extension ambiguity
`input_format` argument solves two different problems right now: it is used to find files in a folder by extension and it is used to select a file reader class. However, this approach has some limitations, since extensions do not necessary correspond to format exactly:
1. Extension can be a bit different for the same format, for example for FITS files it can be `.fits`, `.FITS` or `.FIT`;
2. compressed files would have double extensions, and while we handle it for CSV files, it is also common for FITS files to be gziped, see [Elasticc](https://portal.nersc.gov/cfs/lsst/DESC_TD_PUBLIC/ELASTICC/ELASTICC2_TRAINING_SAMPLE_2/ELASTICC2_TRAIN_02_CART/) for an example;
3. some extensions could correspond to multiple formats, for example Gaia ESCV files have `.csv.gz` extensions, see #225.

delucchi-cmu:
## Change Description

Closes #229. Closes #215.

## Solution Description

This isn't what either issue was asking for, but I think it addresses both of them?

- Removes the `input_format` argument entirely
- the `file_reader` argument can take a `FileReader` instance (current behavior) or can take a string, and it will try to find an appropriate reader (new behavior).
- if providing `input_path`, we will just look for ANY files under that directory. if folks want to be more specific, they can use `glob`, as referenced in documentation.


wilsonbb (approved):
This looks like it addresses my issue, thanks!

src/hipscat_import/catalog/arguments.py
```diff
@@ -74,7 +70,7 @@ class ImportArguments(RuntimeArguments):
     debug_stats_only: bool = False
     """do not perform a map reduce and don't create a new
     catalog. generate the partition info"""
-    file_reader: InputReader | None = None
+    file_reader: InputReader | str | None = "csv"
```
hombit, line 73: Could we make it a required argument?

hombit (approved):
Thank you!

src/hipscat_import/catalog/arguments.py
delucchi-cmu: Done.

PR #236, opened 2024-02-23
Force numpy array for passing to healpy.

### LINKED ISSUE #227: Splitting stage fails when using numpy_nullable backend
**Bug report**

When using `CsvReader(parquet_kwargs={"dtype_backend": "numpy_nullable"})`, splitting fails on this line:
https://github.com/astronomy-commons/hipscat-import/blob/a4cf85a73c3c5e0c429e8140effa83947559239a/src/hipscat_import/catalog/map_reduce.py#L148

It happens because `_iterate_input_file` produces Pandas' `IntegerArray` instead of numpy array of integer type, since `hp.ang2pix` gets Pandas arrays here:
https://github.com/astronomy-commons/hipscat-import/blob/a4cf85a73c3c5e0c429e8140effa83947559239a/src/hipscat_import/catalog/map_reduce.py#L65-L71

It can be solved with replacing `.values` to `.to_numpy(copy=False)`.

**Before submitting**
Please check the following:

- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.

delucchi-cmu:
## Change Description

Closes #227.

### DIFF
--- a/src/hipscat_import/catalog/map_reduce.py
+++ b/src/hipscat_import/catalog/map_reduce.py
@@ -64,8 +64,8 @@ def _iterate_input_file(
             # Set up the pixel data
             mapped_pixels = hp.ang2pix(
                 2**highest_order,
-                data[ra_column].values,
-                data[dec_column].values,
+                data[ra_column].to_numpy(copy=False, dtype=float),
+                data[dec_column].to_numpy(copy=False, dtype=float),
                 lonlat=True,
                 nest=True,
             )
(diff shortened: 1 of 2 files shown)


hombit (commented):
Thank you! I'm not sure if it is required to be changed, but there is also `.values` in the lines 268-269 of the same file

delucchi-cmu:
Added check that the generated hipscat_index values are good (since that's what happening in those lines in question).

hombit (approved):
Looks good!

PR #243, opened 2024-02-28
Sort _hipscat_index column

### LINKED ISSUE #179: Sort by HiPSCat index if use_hipscat_index=True
Currently, this combination of the input parameters make partitions by the HiPSCat index, but doesn't sort by it within each file:
```python
ImportArguments(
    use_hipscat_index=True,
    add_hipscat_index=False,
)
```

It requires an additional `sort_columns=HIPSCAT_ID_COLUMN`. I believe that `use_hipscat_index=True` should trigger sorting, while `sort_columns` should provide additional secondary sorting keys.

### LINKED ISSUE #180: Add pandas index column to metadata
It looks like `ImportArgumetns(use_hipscat_index=True, add_hipscat_index=False)` does not set parquet metadata with pandas index column to be HIPSCAT_ID_COLUMN. I believe that `use_hipscat_index=True` should indicates that user wants it to be set this way.

See also #179

delucchi-cmu:
## Change Description

Closes #179, closes #180

## Solution Description

- Explicitly either creates a new hipscat index, uses an existing hipscat index, or doesn't do anything with hipscat_index. 
- Clumps the functionality under fewer conditionals so the flow of the method is cleaner.
- Sets pandas index to hipscat_index when using existing values and sorts.


src/hipscat_import/catalog/map_reduce.py
```diff
@@ -263,11 +263,28 @@ def reduce_pixel_shards(
...
         )
+        dataframe = dataframe.set_index(HIPSCAT_ID_COLUMN).sort_index()
+
+        # Adjust the schema to make sure that the _hipscat_index will
+        # be saved as a uint64
+        if schema:
+            pandas_index_column = schema.get_field_index("__index_level_0__")
+            if pandas_index_column != -1:
+                schema = schema.remove(pandas_index_column)
+            schema = schema.insert(0, pa.field("_hipscat_index", pa.uint64()))
```
camposandro, line 283:
```suggestion
            schema = schema.insert(0, pa.field(HIPSCAT_ID_COLUMN, pa.uint64()))
```

camposandro (approved):

PR #244, opened 2024-02-28
If not resuming, remove intermediate dir.

### LINKED ISSUE #58: Add --clean option to import when resuming
If you're resuming from a previous interrupted import run, you have to specify --resume to pick up from where you left off.

We could create an additional flag like --clean to remove any intermediate files from previous runs and start fresh.

### LINKED ISSUE #216: ImportArguments.resume=False still sets up a ResumePlan
Even if a user wants to do a short workflow where resume is not needed and therefore `ImportArguments.resume=False`, hipscat-import still sets up an empty ResumePlan and a corresponding directory

This is a fairly minor problem for most workflows, but in the case where a user doesn't have a valid tmp directory configured and then receives errors about constructing the resume plan directory, having to parse those errors related to a feature they disabled can be a source of frustration and confusion

delucchi-cmu:
## Change Description

Closes #58. Closes #216 .

## Solution Description

If the user specifies `resume=False`, then remove the intermediate directory instead of throwing an error. If we are resuming from prior state, give a warning (and catch this warning in testing, as it's a good indication that we've picked up some prior state and we're using it).

### DIFF
--- a/src/hipscat_import/pipeline_resume_plan.py
+++ b/src/hipscat_import/pipeline_resume_plan.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import re
+import warnings
 from dataclasses import dataclass
 from pathlib import Path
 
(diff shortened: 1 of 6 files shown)


wilsonbb (approved):
Thanks!

PR #245, opened 2024-02-29
Improve CSV reading kwargs

### LINKED ISSUE #231: Always use nullable pandas backend
Currently, all nulls in the input CSV data are implicitly converted to the corresponding NA value for a given type, such as NaN for floats. Users may change this with the following code:

```python
CsvReader(
    parquet_kwargs={
        'dtype_backend': 'numpy_nullable', # or 'pyarrow'
    }
)
```

However, it seems like a reasonable default behavior would be to always use one of the nullable dtype backends. "pyarrow" might be more preferable due to its better compatibility with Parquet.

Related to issue #227.

delucchi-cmu:
## Change Description

Improve handling of kwargs for reading parquet schema (to use for column names/types), and for reading CSV file.

### DIFF
--- a/docs/catalogs/arguments.rst
+++ b/docs/catalogs/arguments.rst
@@ -129,7 +129,7 @@ reading CSV, FITS, and Parquet input files, but you can subclass the reader
 type to suit whatever input files you've got.
 
 You only need to provide the ``file_reader`` argument if you are using a custom file reader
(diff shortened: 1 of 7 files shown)


hombit:
I like this PR, but actually #231 was not only about the case when we use `schema_file`, but primary about type derivation, see the code snippet: https://github.com/astronomy-commons/hipscat-import/issues/231#issuecomment-1957752200

hombit (approved):
Looks good

PR #248, opened 2024-03-01
Index generation improvements

### LINKED ISSUE #206: Investigate better process of dropping duplicates in index generation
See also the conversation on https://github.com/astronomy-commons/hipscat-import/pull/202

and the current implementation circa line 57 in /src/hipscat_import/index/map_reduce.py

delucchi-cmu:
## Change Description

Closes #206.

- Performs the index de-duplication when reading from the original leaf file.
- Pass the dask client and use explicitly for the `compute` call.

### DIFF
--- a/src/hipscat_import/index/map_reduce.py
+++ b/src/hipscat_import/index/map_reduce.py
@@ -2,13 +2,31 @@
 
 import dask.dataframe as dd
 import numpy as np
-from dask.distributed import progress, wait
+import pandas as pd
 from hipscat.io import paths
-from hipscat.io.file_io import file_io
 from hipscat.pixel_math.hipscat_id import HIPSCAT_ID_COLUMN
 
 
-def create_index(args):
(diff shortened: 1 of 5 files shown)


src/hipscat_import/index/map_reduce.py
```diff
@@ -36,49 +58,20 @@ def create_index(args):
...
-        #   values are still sorted so it's cheaper.
-        data = (
-            data.repartition(partition_size=1_000_000_000)
-            .reset_index()
-            .drop_duplicates()
-            .set_index(args.indexing_column, sorted=True, partition_size=args.compute_partition_size)
-        )
-    else:
-        data = data.repartition(partition_size=args.compute_partition_size)
+    data = data.repartition(partition_size=args.compute_partition_size)
```
camposandro, line 68: These changes are awesome! I'm not sure we need this line as we're passing the partition size in the `compute_kwargs` below, but I'm assuming it shouldn't be a big deal (?)

camposandro (approved):

PR #260, opened 2024-03-21
Use storage options for input and output

delucchi-cmu:
## Change Description

Addresses problems reading from and writing to cloud buckets during index, margin, and SOAP generation.


camposandro (approved):
Everything looks reasonable! 🙌

Issue #225, opened 2024-02-16
Support ECSV format


hombit:
** Feature request**

It would be great if `hipscat-import` could infer data types from ECSV catalogs, like Gaia DR3. Currently, user is required to provide an empty parquet file with a schema for this catalog.

`astropy` could help with this task. While its ECSV parser looks very slow, we still could use it just for data type inference:
```python
from astropy.io import ascii

empty_astropy_table = ascii.read(gaia_file, format='ecsv', data_end=1)
empty_astropy_table.write('schema.parquet')
```

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #262, opened 2024-03-27
Only read required columns in mapping stage.

delucchi-cmu:
## Change Description

Related to Issue #225. 

I wanted to speed up the ECSV reads by only reading the columns we need during the mapping stage. It's relevant for all kinds of file readers, though.


hombit (approved):
Looks good to me!

PR #279, opened 2024-04-12
Direct to tutorials for public catalogs.

### LINKED ISSUE #192: Add public catalog import tips for gaia
We have dedicated pages for several public catalogs, and some guides around pain points of importing them (https://hipscat-import.readthedocs.io/en/latest/catalogs/public/index.html). Add a page just for gaia.

Original CSV files: http://cdn.gea.esac.esa.int/Gaia/gdr3/gaia_source/

The first 1000 lines are wonky headers that should be ignored on read.

Potentially we might not handle boolean values well on import? Check this out?

Arguments will likely be something like:

```
from hipscat_import.catalog.file_readers import CsvReader

args = ImportArguments(
    output_catalog_name="gaia_csv_test",
    input_path=source_directory,
    file_reader=CsvReader(skiprows=1000)
    input_format="csv.gz",
    ra_column="ra",
    dec_column="dec",
    id_column="source_id",
    output_path=output_path,
    dask_tmp="/epyc/users/ncaplar/", 
    pixel_threshold=500_000,
    highest_healpix_order=6,
    overwrite=True)
```

delucchi-cmu:
Closes #192.

Adds pointer to LSDB tutorials to import and work with a few publicly-available catalogs. 

NB: The links point to stable versions of the LSDB documentation, and the tutorials were recently moved to their new locations. The links won't work until the new version of LSDB is released.

### DIFF
--- a/docs/catalogs/public/index.rst
+++ b/docs/catalogs/public/index.rst
@@ -21,6 +21,8 @@ and provide steps to import those catalogs in case these hints help anyone else.
    sdss
    tic
    zubercal
+   ZTF Bright Transient Survey and New General Catalogue <https://lsdb.readthedocs.io/en/stable/tutorials/ztf_bts-ngc.html>
+   GAIA and DES <https://lsdb.readthedocs.io/en/stable/tutorials/des-gaia.html>
 
 .. tip::
    Want to see more?


hombit (approved):
Links lead to 404, but Sandro said it will be nice after we release LSDB docs...

PR #280, opened 2024-04-17
Resume import with a different HEALPix order

### LINKED ISSUE #261: Handle previous histogram binaries if import arguments change
**Bug report**

In the process of importing a catalog, a user might stop the pipeline before it finishes but after the histogram binaries are generated and stored into disk. If they decide to change the import arguments and re-run the pipeline, it may fail. 

For example, if we change the target HEALPix order, we will stumble upon the following error message:
```
File ~/.conda/envs/python_3.10/lib/python3.10/site-packages/hipscat/pixel_math/partition_stats.py:88, in generate_alignment(histogram, highest_order, lowest_order, threshold)
     60 """Generate alignment from high order pixels to those of equal or lower order
     61 
     62 We may initially find healpix pixels at order 10, but after aggregating up to the pixel
   (...)
     85         exceed threshold.
     86 """
     87 if len(histogram) != hp.order2npix(highest_order):
---> 88     raise ValueError("histogram is not the right size")
     89 if lowest_order > highest_order:
     90     raise ValueError("lowest_order should be less than highest_order")

ValueError: histogram is not the right size
```

Here the importer attempted to resume the pipeline from the same temporary directory and it encountered previously generated histogram binaries which were read to infer the histogram size. Because they differed, the error message is thrown. For more information please have a look at this [workflow](https://github.com/lincc-frameworks/notebooks_lf/blob/main/DELVE_gaia_xmatch/Spectroscopy_failed.ipynb).

We should clean up these files if they are incompatible with the current import configuration or provide the user with a clear message on how to proceed manually.

**Before submitting**
Please check the following:

- [X] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [X] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [X] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

camposandro:
This pull request closes #261. 

- Resolves the redundancy surrounding the `overwrite` / `resume` import arguments, removing `overwrite`  in favor of `resume`. When `resume` is True (default) we attempt to read any intermediate files and continue the import pipeline from an earlier execution. If it is False, we delete any existing intermediate files, we reset the output directory, and start the pipeline from scratch.

- Resolves the issue of loading incompatible histogram binaries whenever trying to resume the import pipeline with a different HEALPix order. We now raise an error so that the user knows that they should set `resume=False`.

- [X] My PR includes a link to the issue that I am addressing

## Code Quality
- [X] I have read the Contribution Guide
- [X] My code follows the code style of this project
- [X] My code builds (or compiles) cleanly without any errors or warnings
- [X] My code contains relevant comments and necessary documentation


camposandro:
I built a small [notebook](https://github.com/lincc-frameworks/notebooks_lf/blob/main/sprints/2024/04_18/resuming-import.ipynb) showcasing the import process for `MagE_hipscat`, the catalog in Julie's workflow. It reproduces the scenario where we run the pipeline at highest pixel order 2, stopping it during the reducing stage, and then proceeding to import the catalog at constant pixel order 3.

src/hipscat_import/catalog/resume_plan.py
```diff
@@ -158,6 +166,16 @@ def write_partial_histogram(cls, tmp_path, mapping_key: str, histogram):
         with open(file_name, "wb+") as file_handle:
             file_handle.write(histogram.data)
 
+    def _reduce_partial_histograms(self, histogram):
```
delucchi-cmu, line 169: The name of this is weird. Maybe `write_combined_histogram`?

src/hipscat_import/catalog/resume_plan.py
```diff
@@ -107,36 +108,43 @@ def read_histogram(self, healpix_order):
...
-                full_histogram = np.add(full_histogram, frombuffer(file_handle.read(), dtype=np.int64))
-
-        file_name = file_io.append_paths_to_pointer(self.tmp_path, self.HISTOGRAM_BINARY_FILE)
-        with open(file_name, "wb+") as file_handle:
-            file_handle.write(full_histogram.data)
-        file_io.remove_directory(
-            file_io.append_paths_to_pointer(self.tmp_path, self.HISTOGRAMS_DIR),
-            ignore_errors=True,
-        )
+                full_histogram = (
```
delucchi-cmu, line 142: Personally, it would be easier to read something like:

```
full_histogram = empty_histogram(size=order)
for file in files:
  with open(blah):
    full_histogram += partial_histogram
```
Is the extra initialization step expensive?

tests/hipscat_import/catalog/test_run_import.py
```diff
@@ -135,6 +134,98 @@ def test_resume_dask_runner(
...
+    assert catalog.on_disk
+    assert catalog.catalog_path == args.catalog_path
+    assert catalog.catalog_info.ra_column == "ra"
+    assert catalog.catalog_info.dec_column == "dec"
+    assert catalog.catalog_info.total_rows == 131
+    assert len(catalog.get_healpix_pixels()) == 4
+
+    # Check that the catalog parquet files exist and
+    # that, collectively, they have the correct set of ids
+    ids = []
```
delucchi-cmu, line 219: we've got a fixture method that does all of this logic: `assert_parquet_file_ids`

delucchi-cmu (commented):
I'm really impressed that you got this to work with just a few changes! And I think this will be much more intuitive for users!

src/hipscat_import/catalog/resume_plan.py
camposandro: I adjusted the code and also added an additional check to make sure the histogram partials have consistent sizes!

tests/hipscat_import/catalog/test_run_import.py
camposandro: I'm generating the catalog at order 1 so I can't call this method just once but I did some polishing!

delucchi-cmu (approved):

PR #288, opened 2024-04-30
Add ECSV reader

### LINKED ISSUE #225: Support ECSV format

delucchi-cmu:
Closes #225 

Adds basic support for an ECSV reader. Note that the astropy ecsv reader does not implement a "fast" reader, so this method is not sped up by underlying C++, and it is not possible to chunk the reads into smaller tables.

See: https://docs.astropy.org/en/stable/io/ascii/read.html#reading-large-tables-in-chunks

### DIFF
--- a/tests/hipscat_import/catalog/test_run_round_trip.py
+++ b/tests/hipscat_import/catalog/test_run_round_trip.py
@@ -12,6 +12,7 @@ import numpy as np
 import numpy.testing as npt
 import pandas as pd
 import pyarrow as pa
+import pyarrow.dataset as pds
 import pyarrow.parquet as pq
 import pytest
 from hipscat.catalog.catalog import Catalog
@@ -484,6 +485,7 @@ def test_gaia_ecsv(
     dask_client,
     formats_dir,
     tmp_path,
+    assert_parquet_file_ids,
 ):
     input_file = os.path.join(formats_dir, "gaia_epoch.ecsv")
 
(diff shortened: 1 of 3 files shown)


tests/hipscat_import/catalog/test_run_round_trip.py
hombit, line 1: I was trying to check if the result catalog has right dtypes. However both pandas and dask.dataframes cannot read the dataset complaining about Norder dtype:
```
E   pyarrow.lib.ArrowTypeError: Unable to merge: Field Norder has incompatible types: uint8 vs int32
```

Could you please modify the test to read the data and check data types?

hombit (changes requested):

tests/hipscat_import/catalog/test_run_round_trip.py
delucchi-cmu: Added tests, though I wasn't able to reproduce the above error.

hombit (approved):
Thank you!

PR #291, opened 2024-05-01
Use preferred string format for globbing.

delucchi-cmu:
See https://github.com/astronomy-commons/lsdb/issues/294

See https://github.com/astronomy-commons/hipscat/pull/265

The `"**"` path part is only supported in the newest fsspec versions as an "entire path component", so it's ok to say `/match/like/**/` but not `/match/like/**.csv`. This should work with older and newer fsspec versions, and so once the change is released for only newer fsspec versions, this package should be ok.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,6 @@ dependencies = [
     "pyyaml",
     "tqdm",
     "numpy",
-    "fsspec <= 2024.2.0", # Remove when pyarrow updates to reflect api changes
 ]
 
 # On a mac, install optional dependencies with `pip install '.[dev]'` (include the single quotes)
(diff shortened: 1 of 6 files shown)


smcguire-cmu (approved):

pyproject.toml
```diff
@@ -25,7 +25,6 @@ dependencies = [
     "pyyaml",
     "tqdm",
     "numpy",
-    "fsspec <= 2024.2.0", # Remove when pyarrow updates to reflect api changes
```
smcguire-cmu, line 28: Actually, is this meant to be removed completely? Do we still want an fsspec dependency?

smcguire-cmu (changes requested):
Just because I hit approve and still have a question sorry

pyproject.toml
delucchi-cmu: `hipscat` contains the fsspec dependency, and hipscat-import shouldn't rely on any fsspec-specific functionality.

smcguire-cmu (approved):

PR #294, opened 2024-05-08
Use scipy.sparse for histogram storage.

### LINKED ISSUE #154: Sparse representation of intermediate histograms on catalog import
When importing a catalog, we store one intermediate healpix histogram per input file as an intermediate file. This uses a very naive np binary array storage, and files can grow large when importing large catalogs at high healpix order.

One alternative is the [healsparse library](https://healsparse.readthedocs.io/en/latest/), (though this introduces additional dependency on healpy).

delucchi-cmu:
Closes #154 using scipy.sparse for representation of healpix histogram.

Using the dense np array on-disk, the intermediate histograms for TIC are 8.5 G. Using sparse array, it's 48 M (around 200X improvement in disk usage). There's no discernible performance penalty.


jeremykubica (approved):
Looks good. I added a bunch of optional comments and thoughts, but nothing blocking.

PR #307, opened 2024-05-14
Quick fixes (no warning, allow . in catalog name).

delucchi-cmu:
Remove the warning from resume functionality. This should be an 'info' level message, but that's more of a logging notion than a true warning.

Allow a catalog name to include a `.`

Address new pylint warnings.


camposandro (approved):
It all looks good!

troyraen (approved):
LGTM!

PR #314, opened 2024-05-17
Spruce up the way we check for original input files.

### LINKED ISSUE #299: Pipeline resume should consider Pathlib path for determining fileset
**Bug report**

If input paths are Pathlib objects, the comparison against previous resumed execution fails.

See `pipeline_resume_plan.check_original_input_paths`

**Before submitting**
Please check the following:

delucchi-cmu:
## Change Description

Closes #299 

## Solution Description

Converts paths to strings before performing comparison, so plain-text strings and Pathlib objects can be compared reasonably. Also, converts both sides into lists for comparisons.

## Code Quality

### Bug Fix Checklist
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### DIFF
--- a/src/hipscat_import/pipeline_resume_plan.py
+++ b/src/hipscat_import/pipeline_resume_plan.py
@@ -172,31 +172,30 @@ class PipelineResumePlan:
         Raises:
             ValueError: if the retrieved file set differs from `input_paths`.
         """
-        unique_file_paths = set(input_paths)
-        unique_file_paths = [str(p) for p in unique_file_paths]
+        input_paths = set(input_paths)
+        input_paths = [str(p) for p in input_paths]
+        input_paths.sort()
 
         original_input_paths = []
 
-        file_path = file_io.append_paths_to_pointer(self.tmp_path, self.ORIGINAL_INPUT_PATHS)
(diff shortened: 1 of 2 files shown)


camposandro:
Looks good to me! 

I just have a small question, should we be writing the set of unique paths, `unique_file_paths`, instead of  `input_paths` to disk? It doesn't really change the behavior in this case since we create a _set_ on read but it might improve readability.

https://github.com/astronomy-commons/hipscat-import/blob/48dc9acc3765954af0a3a46bbbc7b6f0650370c0/src/hipscat_import/pipeline_resume_plan.py#L190-L193

delucchi-cmu:
> Looks good to me!
> 
> I just have a small question, should we be writing the set of unique paths, `unique_file_paths`, instead of `input_paths` to disk? It doesn't really change the behavior in this case since we create a _set_ on read but it might improve readability.
> 
> https://github.com/astronomy-commons/hipscat-import/blob/48dc9acc3765954af0a3a46bbbc7b6f0650370c0/src/hipscat_import/pipeline_resume_plan.py#L190-L193

I think it *would* change the behavior, but only in that it would be more correct. Since we're sorting and de-duping before we check, we should use the results of those operations in the write-to-disk.

camposandro (approved):

PR #318, opened 2024-05-23
Print file name for mapping/splitting failures.

### LINKED ISSUE #301: Print input file for failed map/split stages
**Bug report**

In addition to printing the stack trace, also indicate which file was being processed for the map or split stage that fails. This can help the user track down issues with particular files.

**Before submitting**
Please check the following:

- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.

delucchi-cmu:
## Change Description

Closes #301

Uses dask's distributed print call to log any exception encountered in tasks. This allows for some custom formatting, including the name of the input file, or reference to healpix shard.

## Code Quality

### DIFF
--- a/src/hipscat_import/catalog/map_reduce.py
+++ b/src/hipscat_import/catalog/map_reduce.py
@@ -6,6 +6,7 @@ import healpy as hp
 import numpy as np
 import pyarrow as pa
 import pyarrow.parquet as pq
+from dask.distributed import print
 from hipscat import pixel_math
 from hipscat.io import FilePointer, file_io, paths
 from hipscat.pixel_math.healpix_pixel import HealpixPixel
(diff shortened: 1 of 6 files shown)


src/hipscat_import/catalog/map_reduce.py
```diff
@@ -94,22 +95,29 @@ def map_to_pixels(
...
+                mapped_pixel, count_at_pixel, healpix_order=highest_order
+            )
+            histo.add(partial)
 
-    histo.to_file(ResumePlan.partial_histogram_file(tmp_path=resume_path, mapping_key=mapping_key))
+        histo.to_file(ResumePlan.partial_histogram_file(tmp_path=resume_path, mapping_key=mapping_key))
+    except Exception as exception:  # pylint: disable=broad-exception-caught
+        dask_print("Failed MAPPING stage with file ", input_file)
+        dask_print(exception)
+        raise exception
```
camposandro, line 120: Would it be useful for debugging purposes to also post the worker's address - something like [`dask.distributed.get_worker().address`](https://distributed.dask.org/en/stable/api.html#distributed.get_worker)?

delucchi-cmu: Yes. But a bunch of the unit tests call the method without invoking dask, so it got more complicated =D

src/hipscat_import/pipeline_resume_plan.py
```diff
@@ -220,3 +207,22 @@ def get_pixel_cache_directory(cache_path, pixel: HealpixPixel):
...
+        custom_message (str): some custom message for the task that might help with debugging
+        exception (Exception): error raised in execution of the task.
+    """
+    dask_print(custom_message)
+    try:
+        dask_print("  worker address:", get_worker().address)
+    except Exception:  # pylint: disable=broad-exception-caught
+        pass
+    dask_print(exception)
+    raise exception
```
camposandro, line 228: I think we don't need this instruction here because the exception is already being raised in each task's try catch

delucchi-cmu: Agreed! I thought I had even removed that line.

Thanks for the catch.

camposandro (approved):

PR #345, opened 2024-07-17
Check the total number of rows throughout the catalog import pipeline

### LINKED ISSUE #344: Check the total number of rows throughout the catalog import pipeline
**Feature request**

Allow users to specify the expected number of rows as an import argument. If specified, use this argument to check at various check-points in the pipeline that the number of rows being imported matches the expected number of rows.

Even if the user doesn't specify the expected number of rows, we should be able to determine the number of rows at the end of the mapping stage, and use this number throughout to ensure consistency.

delucchi-cmu:
## Change Description

Closes #344 

## Solution Description

Adds a new argument, `expected_total_rows`, that users can specify if the total number of rows to be imported is known ahead of time. After the mapping stage, we check this against the total number of rows encountered. 

Using the result from the mapping stage, we also check the total for the scheduled reducing stage as well.

## Code Quality


src/hipscat_import/catalog/resume_plan.py
```diff
@@ -249,6 +251,11 @@ def get_alignment_file(
             self.destination_pixel_map = [
                 (order, pix, count) for (order, pix, count) in self.destination_pixel_map if int(count) > 0
             ]
+        total_rows = sum(count for (_, _, count) in self.destination_pixel_map)
+        if total_rows != expected_total_rows:
```
hombit, line 255: Would it fail for the default value (0)?

hombit (approved):
Looks good! I have a minor question

src/hipscat_import/catalog/resume_plan.py
delucchi-cmu: Fair question.

This method doesn't have a default value here. 

In practice, in the pipeline, if they're using the default *argument* of `expected_total_rows=0`, we'll still determine the total rows after the mapping stage. We'll use that value going forward in the pipeline. The call inside run_import.py for `get_alignment_file` passes in the `total_rows` that we got from looking over all of the input files.

troyraen (approved):
LGTM. I think it will also be important to check this value against the number of rows in the parquet files directly at some point. In cases where the import process goes a little funky the parquet files can diverge from what was written in log files, and those are cases where this test can be especially valuable.

delucchi-cmu:
> LGTM. I think it will also be important to check this value against the number of rows in the parquet files directly at some point. In cases where the import process goes a little funky the parquet files can diverge from what was written in log files, and those are cases where this test can be especially valuable.

The method that writes out the parquet metadata will look at all of the footers for the parquet data files. https://github.com/astronomy-commons/hipscat/pull/306 will have that method return the total count. This would also be helpful for a few other import pipelines, where we read the footers AGAIN to calculate the total number of rows written. Once that PR is in, I'll update the code here to add the check.

PR #357, opened 2024-07-26
Provide `drop_empty_siblings` argument

### LINKED ISSUE #269: Highest order when importing should not be needed
https://github.com/astronomy-commons/hipscat-import/blob/9236dab7b045833b3f9af11362daaad2b21626da/src/hipscat_import/catalog/arguments.py#L63

The highest order should be determined from the imported data itself. It seems that now the import fails when importing high-density regions that should have higher orders and we do not specify the max-order, i.e., we use the default value that is too low.

delucchi-cmu:
## Change Description

Closes https://github.com/astronomy-commons/hipscat/issues/302
Closes #269

## Solution Description

Adds `drop_empty_siblings` argument to binning calculation, as added in https://github.com/astronomy-commons/hipscat/pull/304.

Increases the default `highest_order` s.t. users are unlikely to need to change it, and clarifies behavior in the docstrings.

## Code Quality

### DIFF
--- a/src/hipscat_import/catalog/arguments.py
+++ b/src/hipscat_import/catalog/arguments.py
@@ -57,16 +57,21 @@ class ImportArguments(RuntimeArguments):
     a positive number, this will be the order of all final pixels and we
     will not combine pixels according to the threshold"""
     lowest_healpix_order: int = 0
-    """the lowest possible healpix order that we will use for the final 
-    catalog partitioning. setting this higher than 0 will prevent creating
+    """when determining bins for the final partitioning, the lowest possible healpix order 
+    for resulting pixels. setting this higher than 0 will prevent creating
     partitions with a large area on the sky."""
-    highest_healpix_order: int = 7
+    highest_healpix_order: int = 10
     """healpix order to use when mapping. this will
     not necessarily be the order used in the final catalog, as we may combine
     pixels that don't meed the threshold"""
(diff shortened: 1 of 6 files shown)


src/hipscat_import/catalog/arguments.py
```diff
@@ -57,16 +57,21 @@ class ImportArguments(RuntimeArguments):
     a positive number, this will be the order of all final pixels and we
     will not combine pixels according to the threshold"""
     lowest_healpix_order: int = 0
-    """the lowest possible healpix order that we will use for the final 
-    catalog partitioning. setting this higher than 0 will prevent creating
+    """when determining bins for the final partitioning, the lowest possible healpix order 
+    for resulting pixels. setting this higher than 0 will prevent creating
     partitions with a large area on the sky."""
-    highest_healpix_order: int = 7
+    highest_healpix_order: int = 10
```
camposandro, line 63: Should we change this default for LSDB's [`from_dataframe`](https://github.com/astronomy-commons/lsdb/blob/28375b73340bf41c7100ee22a6b91a1edaaba1a0/src/lsdb/loaders/dataframe/from_dataframe.py#L15) to make it consistent?

delucchi-cmu: Ugh. Yes. I'll send a separate PR.

PR #359, opened 2024-07-29
Fix CSV reader from recent changes in hipscat library

delucchi-cmu:
## Change Description

Use appropriate scope for pandas chunk reading generator, with file-ownership in hipscat method.


camposandro (approved):
Thank you!

PR #387, opened 2024-09-12
Replace FilePointer with universal pathlib

delucchi-cmu:
See https://github.com/astronomy-commons/hipscat/pull/336

## Solution Description

Use fsspec's project, [universal-pathlib](https://github.com/fsspec/universal_pathlib) as a path provider. In this way, everywhere we currently pass in a `FilePointer` we can use a `Path` instead, and when we need to create remote connections, use `UPath`.

### DIFF
--- a/docs/catalogs/arguments.rst
+++ b/docs/catalogs/arguments.rst
@@ -169,7 +169,7 @@ You can find the full API documentation for
     )
 
 If you're reading from cloud storage, or otherwise have some filesystem credential
-dict, put those in ``input_storage_options``.
+dict, initialize ``input_file`` using ``universal_pathlib``'s utilities.
 
 Indexed batching strategy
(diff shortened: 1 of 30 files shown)


hombit (commented):
I love how it makes everything much simpler!

There is an inconsistency in `*Arguments`' parameters type-hints: some of them are `UPath`, some are `str` (the same for `List[*]`). Only for these "end-user" interfaces, may I ask to use something more broad, like `str | Path | UPath`?

Do we have a policy that requires any imported package to be an explicit dependency? (it is a good one!) Should we add `universal_pathlib` to `pyproject.toml`?

delucchi-cmu:
> I love how it makes everything much simpler!
> 
> There is an inconsistency in `*Arguments`' parameters type-hints: some of them are `UPath`, some are `str` (the same for `List[*]`). Only for these "end-user" interfaces, may I ask to use something more broad, like `str | Path | UPath`?
> 
> Do we have a policy that requires any imported package to be an explicit dependency? (it is a good one!) Should we add `universal_pathlib` to `pyproject.toml`?

Added a lot more type hints, and direct dependency on `universal_pathlib`.

hombit (approved):
Thank you!

Issue #351, opened 2024-07-24
Complex arrow columns cause an error during the reducing stage


hombit:
**Bug report**

I implemented a custom reader for HSC PDR2 data, which converts a FITS bool-array to `pa.list_(pa.bool_(), 274)`. 

On the reducing stage this ends with an error from pyarrow-pandas conversion layer, originated from `hipscat_import/catalog/map_reduce.py:287, in reduce_pixel_shards()`, where we convert merged pyarrow table to pandas.

I tried to reproduce this problem with a simple pyarrow table including fixed-list column, but this code does work:
```python
import pandas as pd
import pyarrow as pa

a = pa.array([1, 2, 3])
b = pa.array([[True, False, True]] * 3, type=pa.list_(pa.bool_(), 3))
table = pa.Table.from_arrays([a, b], names=['a', 'b'])
df = table.to_pandas(types_mapper=pd.ArrowDtype)
```

Note this types_mapper argument here, it works even without this, but produces object dtype, I believe that we also should use it, but it may be another issue. I also tried to modify `reduce_pixel_shards()` to use this argument, but it produces the same error.

Details are under the spoilers:

<details><summary>Reader implementation</summary>

```python
import numpy as np
import pandas as pd
import pyarrow as pa
from astropy.table import Table
from hipscat_import.catalog.file_readers import FitsReader


class HSCFitsReader(FitsReader):
    """FITS reader that converts ra and dec from radians to degrees"""
    def __init__(self, *args, ra_column, dec_column, **kwargs):
        super().__init__(*args, **kwargs)
        self.ra_column = ra_column
        self.dec_column = dec_column

    def read(self, input_file, read_columns=None):
        # Mostly copy-pasted from the hipscat-import implementation
        self.regular_file_exists(input_file, **self.kwargs)
        table = Table.read(input_file, memmap=True, **self.kwargs)
        if read_columns:
            table.keep_columns(read_columns)
        elif self.column_names:
            table.keep_columns(self.column_names)
        elif self.skip_column_names:
            table.remove_columns(self.skip_column_names)

        total_rows = len(table)
        read_rows = 0

        while read_rows < total_rows:
            chunk = table[read_rows : read_rows + self.chunksize]
            df_chunk = self.astropy_table_to_df(chunk)
            yield df_chunk

            read_rows += self.chunksize

    def astropy_table_to_df(self, table):
        """Data convertion, spoils input table"""
[...]

PR #388, opened 2024-09-16
Support reading a file to pyarrow table in catalog import

delucchi-cmu:
## Change Description

This is related to issue #351, but does not close it.

Allows the InputReader read method to return a pyarrow Table, instead of a Pandas dataframe. Pandas dataframes are converted to pyarrow tables, and all manipulation after that point is based on pyarrow operations. 

This means that tables generated will NOT have the `_hipscat_index` field set as the pandas index. This PR should not be merged until LSDB can support these tables, via fixing https://github.com/astronomy-commons/lsdb/issues/67

### DIFF
--- a/src/hipscat_import/catalog/map_reduce.py
+++ b/src/hipscat_import/catalog/map_reduce.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, Union
 
 import hipscat.pixel_math.healpix_shim as hp
 import numpy as np
+import pandas as pd
 import pyarrow as pa
 import pyarrow.parquet as pq
 from hipscat import pixel_math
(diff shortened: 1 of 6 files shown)


src/hipscat_import/catalog/map_reduce.py
```diff
@@ -50,19 +51,30 @@ def _iterate_input_file(
...
         if use_hipscat_index:
-            if data.index.name == HIPSCAT_ID_COLUMN:
-                mapped_pixels = hipscat_id_to_healpix(data.index, target_order=highest_order)
+            if isinstance(data, pd.DataFrame):
+                if data.index.name == HIPSCAT_ID_COLUMN:
+                    mapped_pixels = hipscat_id_to_healpix(data.index, target_order=highest_order)
+                else:
+                    mapped_pixels = hipscat_id_to_healpix(data[HIPSCAT_ID_COLUMN], target_order=highest_order)
             else:
                 mapped_pixels = hipscat_id_to_healpix(data[HIPSCAT_ID_COLUMN], target_order=highest_order)
```
smcguire-cmu, line 60: I think this could be simplified, with one if statement `if isinstance(data, pd.DataFrame) and data.index.name == HIPSCAT_ID_COLUMN`, so there's not the repeated else statements

smcguire-cmu (approved):
Looks good! Sorry for the late review, I missed that I was assigned this one. One minor comment, but lgtm

delucchi-cmu:
Being replaced by https://github.com/astronomy-commons/hats-import/pull/415

PR #391, opened 2024-09-26
Create HiPS-style properties file

delucchi-cmu:
## Change Description

See https://github.com/astronomy-commons/hipscat/pull/358

This introduces a java-style properties file, similar to the properties file in the HiPS standard.

This fully replaces the catalog_info.json file, and removes the provenance info handling from this library. Individual tools will optionally write out some tool arguments file (see https://github.com/astronomy-commons/hipscat/issues/351).


src/hats_import/index/run_index.py
```diff
@@ -22,14 +22,11 @@ def run(args, client):
...
-            catalog_base_dir=args.catalog_path,
-            dataset_info=index_catalog_info,
-            tool_args=args.provenance_info(),
-        )
+        index_catalog_info = args.to_table_properties(int(rows_written))
         step_progress.update(1)
-        write_metadata.write_catalog_info(catalog_base_dir=args.catalog_path, dataset_info=index_catalog_info)
+        ## TODO - optionally write out arguments file
+        step_progress.update(1)
+        index_catalog_info.to_properties_file(args.catalog_path)
```
camposandro, line 29: Should we move this call to line 26, right after `to_table_properties`, for consistency with the other pipelines? We could then remove this extra `step_progress.update` so that we only have 4 steps, as defined in the `print_progress` arguments.

src/hats_import/margin_cache/margin_cache_arguments.py
```diff
@@ -76,19 +75,18 @@ def _check_arguments(self):
         if margin_pixel_mindist * 60.0 < self.margin_threshold:
             raise ValueError("margin pixels must be larger than margin_threshold")
 
-    def to_catalog_info(self, total_rows) -> MarginCacheCatalogInfo:
+    def to_table_properties(self, total_rows) -> TableProperties:
```
camposandro, line 78:
```suggestion
    def to_table_properties(self, total_rows: int) -> TableProperties:
```

src/hats_import/soap/arguments.py
```diff
@@ -66,21 +65,21 @@ def _check_arguments(self):
         if self.compute_partition_size < 100_000:
             raise ValueError("compute_partition_size must be at least 100_000")
 
-    def to_catalog_info(self, total_rows) -> AssociationCatalogInfo:
+    def to_table_properties(self, total_rows) -> TableProperties:
```
camposandro, line 68:
```suggestion
    def to_table_properties(self, total_rows: int) -> TableProperties:
```

tests/data/small_sky_object_catalog/properties
```diff
@@ -0,0 +1,6 @@
+#utf-8
```
camposandro, line 1: Is this correct? I think I've seen `#HATS catalog` in previous properties files instead.

tests/data/small_sky_source_catalog/properties
```diff
@@ -0,0 +1,6 @@
+#utf-8
```
camposandro, line 1: `#HATS catalog`?

camposandro (commented):
Very minor comments!

tests/data/small_sky_object_catalog/properties
delucchi-cmu: Hah! I just re-ran the `generate_data`, and these were the only two files that changed. Nice catch!

camposandro (approved):

Issue #361, opened 2024-07-31
Catalog import: Option to run a single stage 


delucchi-cmu:
**Feature request**

Add a new argument to specify the stages of the pipeline to run. If not specified, just run the whole thing.

e.g. `run_stages=['mapping', 'splitting']`

**Before submitting**
Please check the following:

- [ ] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [ ] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [ ] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

johnny-gossick:
For static datasets, a hipscat import job may need to be run only once so the total runtime of the job is not a major concern.  However, some datasets will receive regular updates so we may need to run hipscat import on either the full dataset or just the new data (if possible) on a regular schedule.  When dealing with frequently updated catalog data, it will become very important to reduce the runtime of a hipscat job by increasing the parallelism and defining the ideal worker configuration for each stage of the job.  It will also be important to automate these regular hipscat import jobs to reduce manual effort.

Different stages of the pipeline could benefit from having different memory amounts, cpu cores, and numbers of dask workers.  For example, the optimum worker configuration for the mapping stage of an import job with 32 input files might be 32 workers with one core each and X GB of memory because only 32 tasks would be created.  For later stages, there may be much larger task numbers which could benefit from a larger number of dask workers.  In general, some stages may require different amounts of cpu cores or memory.

If hipscat import has the option to choose which stage(s) of the import pipeline to run then we may be able to use [dask gateway](https://gateway.dask.org/cluster-options.html#examples) in a Kubernetes cluster to dynamically spin up different types and numbers of dask workers (up to a set limit) for each stage of the hipscat pipeline.  This could dramatically decrease the total runtime of each import job which would allow us to automate these jobs and run them on a regular schedule.

delucchi-cmu:
@troyraen I think this would have also helped you in the past with performing manual verification in-between stages of the import pipeline.

troyraen:
[...]

PR #406, opened 2024-10-15
Insert dataset dir, and use general ra/dec columns

delucchi-cmu:
## Change Description

See https://github.com/astronomy-commons/hipscat/pull/377

Adds a dataset directory that contains all parquet files (leaf parquet data files, and root-level metadata files).

Other changes included:

- sneaks in change to progress bar
- give more helpful error message when row counts don't match in conversion
- re-generate unit test data with these changes included.


src/hats_import/hipscat_conversion/run_conversion.py
```diff
@@ -92,7 +92,11 @@ def run(args: ConversionArguments, client):
         simple_progress_bar=args.simple_progress_bar,
     ) as step_progress:
         total_rows = parquet_metadata.write_parquet_metadata(args.catalog_path)
-        assert total_rows == properties.total_rows
+        if total_rows != properties.total_rows:
+            raise ValueError(
+                f"Unexepcted number of rows (original: {properties.total_rows}"
```
camposandro, line 97:
```suggestion
                f"Unexpected number of rows (original: {properties.total_rows}"
```

tests/data/small_sky_object_catalog/properties
```diff
@@ -2,13 +2,13 @@
...
+hats_col_dec=dec
 hats_max_rows=1000000
 hats_order=0
 moc_sky_fraction=0.08333
-hats_builder=hats-import v0.3.6.dev22+g6346e4d
-hats_creation_date=2024-09-30T13\:52UTC
-hats_estsize=121
+hats_builder=hats-import v0.3.6.dev26+g40366b4
+hats_creation_date=2024-10-11T15\:02UTC
+hats_estsize=74
```
camposandro, line 12: Is this estimated size smaller because it's now only accounting for the parquet files?

delucchi-cmu: I suspect that it's due to hats PR [#361](https://github.com/astronomy-commons/hats/pull/361) - this switched the pointmap.fits file to store dtype=np.int32 instead of int64. which would cut the size of that file by about 50%, and most of the storage for very-small-catalogs is taken up by that one file.

PR #418, opened 2024-10-22
Keep track of expected number of margin rows written.

### LINKED ISSUE #258: Confusing error behavior if margin generation results in empty catalog
**Bug report**

If you set a margin that is too low to result in any objects falling in the margins, the pipeline fails on the `parquet_metadata.write_parquet_metadata` call with varying error messages, depending on your file system.

On a local filesystem:

```
pyarrow.lib.ArrowInvalid: ParquetDatasetFactory must contain a schema with at least one column
```

For azure:

```
FileNotFoundError: abfs://hipscat/pytests/data/small_sky_order1_margin
```

If there are no points found in the first mapping step, then we should error out of the pipeline early, stating that no margin objects were found and they should increase the margin.

**Before submitting**
Please check the following:

delucchi-cmu:
Closes #258.

As a bonus, we can track the number of rows that are expected from the mapping stage, and we can double-check total after reducing stage.


src/hats_import/margin_cache/margin_cache_resume_plan.py
```diff
@@ -163,3 +180,10 @@ def _find_partition_margin_pixel_pairs(combined_pixels, margin_order):
         columns=["partition_order", "partition_pixel", "margin_pixel"],
     ).sort_values("margin_pixel")
     return margin_pairs_df
+
+
+def _marker_value_to_int(marker_value: List[str]) -> int:
```
gitosaurus, line 185:
```suggestion
def _marker_value_to_int(marker_value: list[str]) -> int:
```

src/hats_import/pipeline_resume_plan.py
```diff
@@ -85,12 +85,46 @@ def touch_key_done_file(cls, tmp_path, stage_name, key):
...
+        )
+
+    def read_markers(self, stage_name):
+        """Inspect the stage's directory of marker files, fetching the key value pairs
+        from marker file names and contents.
+
+        Args:
+            stage_name(str): name of the stage (e.g. mapping, reducing)
+        Return:
+            List[str] - all keys found in done directory
```
gitosaurus, line 117:
```suggestion
            list[str] - all keys found in done directory
```

src/hats_import/pipeline_resume_plan.py
```diff
@@ -85,12 +85,46 @@ def touch_key_done_file(cls, tmp_path, stage_name, key):
...
+            tmp_path: where to write pipeline intermediate files
+            stage_name(str): name of the stage (e.g. mapping, reducing)
+            key (str): unique string for each task (e.g. "map_57")
+            value (str): value for the marker.
+        """
+        file_io.write_string_to_file(
+            file_io.append_paths_to_pointer(tmp_path, stage_name, f"{key}_done"), value
+        )
+
+    def read_markers(self, stage_name):
```
gitosaurus, line 110:
```suggestion
    def read_markers(self, stage_name: str) -> list[str]:
```

src/hats_import/margin_cache/margin_cache.py
```diff
@@ -39,6 +39,11 @@ def generate_margin_cache(args, client):
             )
         resume_plan.wait_for_mapping(futures)
 
+    with resume_plan.print_progress(total=1, stage_name="Binning") as step_progress:
+        total_rows = resume_plan.get_mapping_total()
+        if not total_rows:
+            raise ValueError("Margin cache contains no rows. Increase margin size and re-run.")
```
gitosaurus, line 45:
```suggestion
            raise ValueError("Margin cache contains no rows.  Increase margin size and re-run.")
```
Such a nitpick.  Sorry!  Just the way I was raised when writing with monospaced fonts for monospaced display.  🤦  Feel free to ignore, particularly if it contravenes existing message style in this project.

tests/hats_import/margin_cache/test_margin_cache.py
```diff
@@ -97,3 +97,19 @@ def test_margin_cache_gen_negative_pixels(small_sky_source_catalog, tmp_path, da
...
+    args = MarginCacheArguments(
+        margin_threshold=10.0,
+        input_catalog_path=small_sky_object_catalog,
+        output_path=tmp_path,
+        output_artifact_name="catalog_cache",
+        margin_order=8,
+        progress_bar=False,
+    )
+
+    with pytest.raises(ValueError, match="Margin cache contains no rows"):
```
gitosaurus, line 114: 😎 I didn't know about the `match=` keyword there; very cool

gitosaurus (approved):
Looks good!  Tiny stylistic comments only.

src/hats_import/margin_cache/margin_cache.py
delucchi-cmu: Other messages and documentation in the project use a single space. This is now the recommendation of the Chicago Manual of Style, and several other style guides. Within the team, we defer to the Rubin DM developer style guide, which has some opinions, [at least for docstrings](https://developer.lsst.io/python/style.html#sentences-in-comments-should-not-be-separated-by-double-spaces), but also prefer single-space after sentence.

PR #421, opened 2024-10-29
Update dependencies

delucchi-cmu:
- pyyaml was unused
- scipy is no longer a direct dependency (https://github.com/astronomy-commons/hats-import/pull/405)
- ipykernel is a nice-to-have dependency for folks running inside a jupyter notebook
- but ipywidgets is an even-nicer-to-have dependency
- extras handles the nice-to-haves, including the full suite of fsspec implementations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,12 +18,9 @@ dependencies = [
     "dask[complete]>=2024.3.0", # Includes dask expressions.
     "deprecated",
     "hats >=0.4",
-    "ipykernel", # Support for Jupyter notebooks
     "numpy",
     "pandas",
     "pyarrow",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -41,7 +38,11 @@ dev = [
     "pytest-cov",
     "pytest-timeout",
     "ray", # Used for dask-on-ray testing.
-    "types-PyYAML", # type stubs for pyyaml
+]
+extra = [
```
hombit, line 42:
```suggestion
full = [
```

hombit (approved):

hombit:
Could you please also update Readme and RTD about this new extras?

PR #436, opened 2024-11-18
Use different methods for healpix manipulation.

delucchi-cmu:
See also 

https://github.com/astronomy-commons/hats/pull/422


gitosaurus (approved):
LGTM!

PR #440, opened 2024-11-25
Remove default healpy dependency.

delucchi-cmu:
See https://github.com/astronomy-commons/hats/pull/433

We still need the `healpy` library to occasionally read older FITS files that may have been written in RING pixelization. Marked as optional dependency, and tests can still run against it if healpy is installed.


gitosaurus (approved):
Looks great!

PR #443, opened 2024-11-26
Force re-creation of margin catalog.

delucchi-cmu:
See https://github.com/astronomy-commons/hats-import/issues/434

Existing margins may have missed many points when generated. Don't allow blind conversion from hipscat -> hats.


gitosaurus (approved):
LGTM

PR #456, opened 2024-12-10
Perform extra tests against empty files.

delucchi-cmu:
Just checks that empty input files (both regular input, and indexed input) don't cause issues with the mapping or binning stages.


gitosaurus (approved):
LGTM!

PR #464, opened 2024-12-28
Avoid overwriting valid catalogs.

### LINKED ISSUE #459: Avoid overwriting a valid HATS catalog during import
**Feature request**

It is possible to check whether a given path points to a valid catalog, like this:

```python
from hats.io.validation import is_valid_catalog
is_valid_catalog(potential_path)
```

If this check returns `True` when `hats_import.pipeline` begins, it should refuse to write to that directory, and clearly state to the user that it won't overwrite a valid catalog.  This will both alert the user to what they need to do (find a fresh path or move the existing one), and also prevent the accidental overwrite of good data.

When the pipeline is set up to resume (as it is by default), it is easy to run across this condition, since extra warning detritus from dask shutdown can hide the fact that an incrementally proceeding import has finally completed; the next attempt to resume will end up restarting the import, which is almost never what is desired.

An `overwrite=` flag could be introduced to permit catalog overwrites, but its utility seems limited.

A good place to put this check, at least to start with, is `src/hats_import/catalog/run_import.py`, in the `run()` function.  It could be put higher in the call stack, in `src/hats_import/pipeline.py`, in the `pipeline_with_client` function; this would cause it to apply to every pipeline, so that every pipeline would avoid overwriting a valid catalog.

**Before submitting**
Please check the following:

- [X] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [X] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [X] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

gitosaurus:
Strong refusal without options, since it's easier and safer to move the catalog out of the way.

## Change Description

Before beginning a catalog import or generating a margin cache, check whether the output artifact already contains a valid catalog, and if so, stop by raising a `ValueError`.

Closes #459 .

- [X] My PR includes a link to the issue that I am addressing

## Code Quality
- [X] I have read the [Contribution Guide](https://hats-import.readthedocs.io/en/stable/guide/contributing.html) and [LINCC Frameworks Code of Conduct](https://lsstdiscoveryalliance.org/programs/lincc-frameworks/code-conduct/)
- [X] My code follows the code style of this project
- [X] My code builds (or compiles) cleanly without any errors or warnings
- [X] My code contains relevant comments and necessary documentation

### Bug Fix Checklist

Tested this end-to-end on epyc, and verified it, and added unit tests to verify argument validation.

- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### DIFF
--- a/src/hats_import/catalog/run_import.py
+++ b/src/hats_import/catalog/run_import.py
@@ -5,8 +5,8 @@ The actual logic of the map reduce is in the `map_reduce.py` file.
 """
 
 import os
-from pathlib import Path
 import pickle
+from pathlib import Path
 
 import hats.io.file_io as io
 from hats.catalog import PartitionInfo
--- a/src/hats_import/catalog/run_import.py
+++ b/src/hats_import/catalog/run_import.py
@@ -19,17 +19,26 @@ from hats_import.catalog.arguments import ImportArguments
 from hats_import.catalog.resume_plan import ResumePlan
 
 
-def run(args, client):
-    """Run catalog creation pipeline."""
+def _validate_arguments(args):
+    """
+    Verify that the args for run are valid: they exist, are of the appropriate type,
+    and do not specify an output which is a valid catalog.
+
+    Raises ValueError if they are invalid.
+    """
     if not args:
         raise ValueError("args is required and should be type ImportArguments")
     if not isinstance(args, ImportArguments):
         raise ValueError("args must be type ImportArguments")
 
-    # Verify that the planned output path is not occupied by a valid catalog
     potential_path = Path(args.output_path) / args.output_artifact_name
     if is_valid_catalog(potential_path):
         raise ValueError(f"Output path {potential_path} already contains a valid catalog")
+
+
+def run(args, client):
+    """Run catalog creation pipeline."""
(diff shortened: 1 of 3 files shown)


maxwest-uw:
Should this be using the `strict` option for `is_valid_catalog` ? naively it seems like the non-strict option only checks if there are valid partition and metadata files (which might cause problems in the cause where an import was paused/interrupted). Otherwise this looks good to me !

gitosaurus:
> Should this be using the `strict` option for `is_valid_catalog` ? naively it seems like the non-strict option only checks if there are valid partition and metadata files (which might cause problems in the cause where an import was paused/interrupted). Otherwise this looks good to me !

The main reason is performance: `strict=False` (the default) takes ~3ms to run, while `strict=True` takes ~25s.  Moreover, `strict=False` will still correctly reject in-progress catalogs as invalid, in my testing.

maxwest-uw (approved):

PR #466, opened 2025-01-06
Move existing catalog check.

delucchi-cmu:
## Change Description

See also #459 and #464

Moves check for existing valid catalog to earlier in the argument checks.


gitosaurus (approved):
LGTM! And thanks.

PR #468, opened 2025-01-07
Address slow list operations in progress tracking.

delucchi-cmu:
The current strategy of reading strings from file names, passing along those strings, and manually performing `not in` against constructed strings was taking foooorrrrreeeevvvveeeerrrrr in cases where there are 100k+ stages. This streamlines those operations, using integers and set operations where possible.

### DIFF
--- a/src/hats_import/catalog/resume_plan.py
+++ b/src/hats_import/catalog/resume_plan.py
@@ -302,9 +302,9 @@ class ResumePlan(PipelineResumePlan):
                 alignment = pickle.load(pickle_file)
             non_none_elements = alignment[alignment != np.array(None)]
             pixel_list = np.unique(non_none_elements)
-            self.destination_pixel_map = dict(
-                [(HealpixPixel(order, pix), count) for (order, pix, count) in pixel_list if int(count) > 0]
-            )
+            self.destination_pixel_map = {
+                HealpixPixel(order, pix): count for (order, pix, count) in pixel_list if int(count) > 0
+            }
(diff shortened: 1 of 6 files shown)


src/hats_import/catalog/resume_plan.py
```diff
@@ -144,13 +145,12 @@ def get_remaining_map_keys(self):
...
         """
-        prefix = file_io.append_paths_to_pointer(self.tmp_path, self.HISTOGRAMS_DIR)
-        mapped_keys = self.get_keys_from_file_names(prefix, ".npz")
-        return [
-            (f"map_{i}", file_path)
-            for i, file_path in enumerate(self.input_paths)
-            if f"map_{i}" not in mapped_keys
+        prefix = file_io.get_upath(self.tmp_path) / self.HISTOGRAMS_DIR
+        done_indexes = [
+            int(re.match(r"map_(.*).npz", str(path.name)).group(1)) for path in prefix.glob("*.npz")
```
gitosaurus, line 150:
```suggestion
            int(re.match(r"map_(\d+).npz", path.name).group(1)) for path in prefix.glob("*.npz")
```

Also: here and elsewhere, using `re.compile` to save patterns to module variables will improve performance further, perhaps enough to notice.

gitosaurus (commented):
Looks good!  Most of my comments are quite optional, but I believe that compiling the regexes is worth doing if we know the data volume is large here.

gitosaurus (approved):
LGTM!

PR #474, opened 2025-01-21
Add more detail about available HATS properties on import.

delucchi-cmu:
Expose documentation about the available automatic and user-provided properties of HATS catalogs.


OliviaLynn (approved):
Looks good!

PR #476, opened 2025-01-22
Suggested changes from science homework.

delucchi-cmu:
See: https://github.com/astronomy-commons/lsdb/issues/541

- Give (only slightly) more friendly error message when some stages of the import task fail
- Clarify that the "public catalogs" page is just tips for various catalogs
- Provide link to data.lsdb.io
- Fix doc links for import notebooks


gitosaurus (approved):
LGTM!

PR #477, opened 2025-01-22
Add verbosity option, convenience pass/fail, and user documentation

delucchi-cmu:
Adds some features and documentation that was useful when building a demo of this pipeline.

- an option to turn off the standard out status reporting
- a simple property to determine if all of the tests passed
- user documentation and links to alternative verification recipes.


hombit (approved):
Looks cool

PR #503, opened 2025-03-05
Add hats reimport function

### LINKED ISSUE #441: Hats reimport pipeline
**Feature request**

It would be useful to have a HATS reimport pipeline to simplify tasks such as re-partitioning HATS catalogs. This could be implemented as a thin wrapper around `ImportArguments`, which would handle file paths, the format (parquet), `use/add_healpix_29`, and default to using the same catalog name.

I had to write something like this to do that, and it took me a while to figure it out.

```python
catalog = read_hats(input_catalog_path)

column_names = catalog.schema.names
column_names.remove('Dir')
column_names.remove('Norder')
column_names.remove('Npix')

args = ImportArguments(
    ra_column=catalog.catalog_info.ra_column,
    dec_column=catalog.catalog_info.dec_column,
    input_path=input_catalog_path / "dataset",
    file_reader=ParquetReader(column_names= column_names),
    output_artifact_name=catalog.catalog_name,
    output_path=output_dir,
    pixel_threshold=100_000,
    use_healpix_29=True,
    add_healpix_29=False,
)
```

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [ ] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

smcguire-cmu:
Adds a class method to `ImportArguments` to generate arguments from an existing hats catalog.

Fixes #441


tests/hats_import/catalog/test_reimport_hats.py
```diff
@@ -0,0 +1,43 @@
+import hats
+import pytest
+from hats.io.paths import DATASET_DIR
+
+from hats_import.catalog import ImportArguments
+from hats_import.catalog.file_readers import ParquetReader
+
+
+def test_reimport_arguments(tmp_path, small_sky_object_catalog):
```
delucchi-cmu, line 9: Can you please add a test that executes the import pipeline with constructed arguments, and verify that the resulting catalog is valid?

src/hats_import/catalog/arguments.py
```diff
@@ -137,6 +140,46 @@ def to_table_properties(
...
+        if column_names is not None:
+            column_names.remove(PARTITION_ORDER)
+            column_names.remove(PARTITION_DIR)
+            column_names.remove(PARTITION_PIXEL)
+
+        import_args = {
+            "catalog_type": catalog.catalog_info.catalog_type,
+            "ra_column": catalog.catalog_info.ra_column,
+            "dec_column": catalog.catalog_info.dec_column,
+            "input_path": path / DATASET_DIR,
```
delucchi-cmu, line 172: I think this will also include `_metadata` and `_common_metadata` files.

src/hats_import/catalog/arguments.py
```diff
@@ -137,6 +140,46 @@ def to_table_properties(
...
+            "file_reader": ParquetReader(column_names=column_names),
+            "output_artifact_name": catalog.catalog_name,
+            "output_path": output_dir,
+            "use_healpix_29": True,
+            "add_healpix_29": False,
```
delucchi-cmu, line 177: There are many other possible [user-provided](https://hats-import.readthedocs.io/en/latest/catalogs/properties.html#user-provided-properties) properties values that might be present in the original catalog's `properties` file that we should attempt to carry forward using the `addl_hats_properties` arg.

src/hats_import/catalog/arguments.py
```diff
@@ -137,6 +140,46 @@ def to_table_properties(
...
+        }
```
delucchi-cmu, line 178: Might be nice to go ahead and specify the `_common_metadata` of the original catalog via `use_schema_file` argument.

src/hats_import/catalog/arguments.py
```diff
@@ -137,6 +140,46 @@ def to_table_properties(
...
+
```
delucchi-cmu, line 179: Set `expected_total_rows` to original catalogs number of rows, just for a little added security.

tests/hats_import/catalog/test_reimport_hats.py
```diff
@@ -0,0 +1,43 @@
...
+    assert args.ra_column == catalog.catalog_info.ra_column
+    assert args.dec_column == catalog.catalog_info.dec_column
+    assert args.input_path == catalog.catalog_base_dir / DATASET_DIR
+    assert isinstance(args.file_reader, ParquetReader)
+    assert args.output_artifact_name == catalog.catalog_name
+    assert args.use_healpix_29
+    assert not args.add_healpix_29
+
+
+def test_reimport_arguments_extra_kwargs(tmp_path, small_sky_object_catalog):
```
delucchi-cmu, line 22: What if I wanted to re-import and change the name? Instead of re-import to a new parent directory? e.g.

```
_ my_catalogs/
  |_  generated_via_to_hats/
  |   ...
  |_  reimported_via_reimport_from_hats/
  |   ...
```

Does that work? How does that work? Can we add user documentation for it?

delucchi-cmu (commented):
Let's also add a page of documentation to RTD on running such a pipeline.

src/hats_import/catalog/arguments.py
```diff
@@ -137,6 +140,46 @@ def to_table_properties(
         } | self.extra_property_dict()
         return TableProperties(**info)
 
+    @classmethod
+    def reimport_from_hats(
+        cls, path: str | Path | UPath, output_dir: str | Path | UPath, **kwargs
```
hombit, line 145: I don't like `**kwargs` here, because it is really hard to find what goes there in the API reference.

hombit (changes requested):
Thank you!

Could you please add user-level docs, because otherwise I have nothing to copy-paste from!

src/hats_import/catalog/arguments.py
smcguire-cmu: I could add all the properties from the arguments as individual kwargs, but it'd be hard to keep those updated. Would a link to the args in the docstring work for you? Or I could add the commonly used ones (output_artifact_name, threshold etc.) and keep the kwargs around?

tests/hats_import/catalog/test_reimport_hats.py
smcguire-cmu: It works I think! If the user passes in `output_artifact_name` as a kwarg it will be updated to whatever they update it to. But having that in the docs is a good idea.

src/hats_import/catalog/arguments.py
smcguire-cmu: I tried this and it actually causes an error because of the Norder, Dir and Npix columns. Since we add those in the import pipeline we have to remove those from the columns of the File Reader. But I don't think there's a way to specify removing columns from the schema file to use. It might be worth adding the option to specify a schema as an import argument and not just a schema file, but that might be out of scope of this PR.

src/hats_import/catalog/arguments.py
hombit: The docstring link is fine!

src/hats_import/catalog/arguments.py
delucchi-cmu: There's a complementary method, `explicit_dict`, on the TableProperties object. You could create one that just returns the explicitly provided `__pydantic_extra__` fields.

src/hats_import/catalog/arguments.py
smcguire-cmu: Well this problem no longer exists!

src/hats_import/catalog/arguments.py
```diff
@@ -134,9 +138,53 @@ def to_table_properties(
...
+            A ImportArguments object with the arguments from the existing catalog, and any updates from kwargs
+        """
+
+        path = get_upath(path)
+
+        catalog = hats.read_hats(path)
+
+        column_names = catalog.schema.names if catalog.schema is not None else None
+
+        in_file_paths = glob.glob(
```
delucchi-cmu, line 166: Paths should still be instances of UPath. You may want to go ahead and add a test case for this inside hats-cloudtests, as I suspect this may cause problems if reading from non-local file systems.

src/hats_import/catalog/arguments.py
```diff
@@ -2,14 +2,18 @@
...
 
+import hats
 from hats.catalog import TableProperties
+from hats.io.file_io import get_upath
+from hats.io.paths import DATASET_DIR, PARTITION_ORDER
 from hats.pixel_math import spatial_index
 from upath import UPath
 
-from hats_import.catalog.file_readers import InputReader, get_file_reader
+from hats_import.catalog.file_readers import InputReader, ParquetReader, get_file_reader
```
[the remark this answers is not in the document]
delucchi-cmu: This will be marginally more performant (I think).

tests/hats_import/catalog/test_reimport_hats.py
```diff
@@ -0,0 +1,119 @@
...
+    assert not args.add_healpix_29
+
+
+def test_reimport_arguments_wrong_dir(tmp_path):
+    wrong_input_path = tmp_path / "nonsense"
+    with pytest.raises(FileNotFoundError):
+        ImportArguments.reimport_from_hats(wrong_input_path, tmp_path)
+
+
+@pytest.mark.timeout(30)
```
delucchi-cmu, line 78: Why does this test need a larger timeout than the pipeline that originated the catalog?

src/hats_import/catalog/arguments.py
```diff
@@ -134,9 +138,53 @@ def to_table_properties(
...
+            "dec_column": catalog.catalog_info.dec_column,
+            "input_file_list": in_file_paths,
+            "file_reader": ParquetReader(column_names=column_names),
+            "output_artifact_name": catalog.catalog_name,
+            "output_path": output_dir,
+            "use_healpix_29": True,
+            "add_healpix_29": False,
+            "use_schema_file": hats.io.paths.get_common_metadata_pointer(catalog.catalog_base_dir),
+            "expected_total_rows": catalog.catalog_info.total_rows,
+            "addl_hats_properties": catalog.catalog_info.extra_dict(by_alias=True),
```
delucchi-cmu, line 182: There are other properties that are somewhere between "extra" and the ones called out here. e.g. `hats_cols_default` and `hats_npix_suffix`.

Will those get carried along? Can we add some default columns to the `small_sky_object_catalog` so that can be tested?

src/hats_import/catalog/arguments.py
```diff
@@ -134,9 +138,53 @@ def to_table_properties(
...
+        """
+
+        path = get_upath(path)
+
+        catalog = hats.read_hats(path)
+
+        column_names = catalog.schema.names if catalog.schema is not None else None
+
+        in_file_paths = glob.glob(
+            str(path / DATASET_DIR / f"{PARTITION_ORDER}*/**/*.parquet"), recursive=True
```
delucchi-cmu, line 167: Consider using the value of `catalog.catalog_info.hats_npix_suffix` here.

tests/hats_import/catalog/test_reimport_hats.py
```diff
@@ -0,0 +1,119 @@
...
+@pytest.mark.dask
+def test_run_reimport(
+    dask_client,
+    small_sky_object_catalog,
+    tmp_path,
+):
+    output_name = "small_sky_higher_order"
+    pixel_thresh = 100
+    args = ImportArguments.reimport_from_hats(
+        small_sky_object_catalog, tmp_path, pixel_threshold=pixel_thresh, output_artifact_name=output_name
```
delucchi-cmu, line 88:
```suggestion
        small_sky_object_catalog, tmp_path, pixel_threshold=pixel_thresh, output_artifact_name=output_name, addl_hats_properties={"obs_regime": "Optical"},
```

delucchi-cmu: That's my personal favorite additional info parameter for unit tests

tests/hats_import/catalog/test_reimport_hats.py
```diff
@@ -0,0 +1,119 @@
...
+
+    old_cat = read_hats(small_sky_object_catalog)
+
+    # Check that the catalog metadata file exists
+    catalog = read_hats(args.catalog_path)
+    assert catalog.on_disk
+    assert catalog.catalog_path == args.catalog_path
+    assert catalog.catalog_info.ra_column == old_cat.catalog_info.ra_column
+    assert catalog.catalog_info.dec_column == old_cat.catalog_info.dec_column
+    assert catalog.catalog_info.total_rows == old_cat.catalog_info.total_rows
```
delucchi-cmu, line 101:
```suggestion
    assert catalog.catalog_info.total_rows == old_cat.catalog_info.total_rows
    assert catalog_info.__pydantic_extra__["obs_regime"] == "Optical"
```

hombit (approved):
Thank you

tests/hats_import/catalog/test_reimport_hats.py
smcguire-cmu: I wasn't setting the maximum healpix order so it was doing it at the default, that's fixed now! Is there anywhere in the properties that could be a better default for this? Or should we just leave it to the user like with the regular import arguments?

src/hats_import/catalog/arguments.py
smcguire-cmu: I've updated it so they're now included. I couldn't see any other properties like this, are there any you can think of?

tests/hats_import/catalog/test_reimport_hats.py
delucchi-cmu: Probably best to leave it to the user. A primary motivation for this pipeline would be to re-partition and likely get a different order from the original, so I don't think we should use hints from the original catalog here.

PR #506, opened 2025-03-18
Don't generate Norder/Dir/Npix columns at import.

delucchi-cmu:
No longer adds the Norder/Dir/Npix columns to the parquet when the catalog is imported.

This will also suppress these columns when supporting tables (margins, indexes) are created. Regenerates test data to match expectations.

See https://github.com/astronomy-commons/hats/issues/367


camposandro (approved):
It's great to finally get rid of these! I see that you're already suppressing those on the LSDB side as well - thank you!

Issue #517, opened 2025-04-21
Change default Parquet compression for import, and make it configurable


gitosaurus:
**Feature request**
Per the survey done as part of https://github.com/astronomy-commons/lsdb/issues/695, change the default Parquet compression scheme from `SNAPPY` to `ZSTD` level `15`.  This should be no more than a parameter change to `pq.write_table` and its equivalents.

Besides making this change to the default, also allow users to specify the compression level in the `ImportArguments`, so that in case there is a better scheme for a particular catalog, it can be used.

It would also be good to limit the allowed compression schemes to those which are supported by CUDA GPU libraries, and also to non-deprecated [settings](https://parquet.apache.org/docs/file-format/data-pages/compression/) (i.e., not `LZ4`).

**Before submitting**
Please check the following:

- [X] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [ ] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [X] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

delucchi-cmu:
See https://github.com/astronomy-commons/hats-import/issues/516

Issue #513, opened 2025-04-10
Import with (configurable) row group splitting


delucchi-cmu:
**Feature request**

For files >100MB, having smaller row groups can provide a read-time speed-up. However, the way in which the row groups are split would vary according to some information about the survey (i.e. ztf and gaia have different "best" strategies).

Note that this implies that row group information should likely NOT appear in the IVOA standard, as there is no single-best-way to generate row groups for all catalogs.

Based on zoom-hack, we have the following pseudo-code:

```
args -= ImportArguments(
	...
	row_group_kwargs={subtile_order_delta: 2},
or
	row_group_kwargs={sort_column:"id", num_rows: 50_000},
)
```
and
```
def reduce_pixels (
	...
	row_group_kwargs,
):

	table = ...
	tables = _split_to_row_groups(table, row_group_kwargs)

	## extra credit
       with pq.ParquetWriter(output_file, tables[0].schema, 
filesystem=file_system) as writer:
            for table in tables:
                writer.write_table(table,
 sort_columns=<the sort column used for row groups>)

def _split_to_row_groups(table, row_group_kwargs):
## this can be as wonky as we need
	if "sort_column" in row_group_kwargs:
		## do it that way
	if "subtile_order_delta" in row_group_kwargs:
		## do it the other way
```

PR #519, opened 2025-04-22
Pass additional kwargs through import

### LINKED ISSUE #516: Add write_table_kwargs
**Feature request**

Based on the results of analysis for https://github.com/astronomy-commons/lsdb/issues/695, it may be useful for other compression schemes to be used for data storage.

This is set via one or two arguments to `pyarrow.parquet.write_table`. Additional arguments are set on this call that a user may want to specify, so we should enable passing various arguments, e.g. 

```
args = ImportArguments(
  ...
  write_table_kwargs={"compression": "zstd",
                      "compression_level": 15,
                      "store_decimal_as_integer": True}
)
```

See https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html

delucchi-cmu:
Closes #516, #517.
Related to #513.

Passes additional kwarg dictionaries through the import pipeline to use when writing the final parquet tables. 
- `write_table_kwargs` set is passed directly to the `write_table` call, and can specify values for compression
- `row_group_kwargs` defines the row group splitting strategy to use for the parquet files. currently the only strategy supported is the current sort column.

### DIFF
--- a/src/hats_import/catalog/arguments.py
+++ b/src/hats_import/catalog/arguments.py
@@ -130,14 +130,13 @@ class ImportArguments(RuntimeArguments):
         # Basic checks complete - make more checks and create directories where necessary
         self.input_paths = find_input_paths(self.input_path, "**/*.*", self.input_file_list)
 
-        if self.write_table_kwargs:
-            # If user has provided some args, let's make sure they're good.
-            if "compression" in self.write_table_kwargs:
-                compression_type = self.write_table_kwargs["compression"]
-                ## TODO!
-                print(compression_type)
-        else:
-            self.write_table_kwargs = {"compression": "ZSTD", "compression_level": 15}
+        if self.write_table_kwargs is None:
+            self.write_table_kwargs = {}
+        if "compression" not in self.write_table_kwargs:
+            self.write_table_kwargs = self.write_table_kwargs | {
+                "compression": "ZSTD",
+                "compression_level": 15,
+            }
 
     def to_table_properties(
         self, total_rows: int, highest_order: int, moc_sky_fraction: float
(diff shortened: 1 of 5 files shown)


src/hats_import/catalog/map_reduce.py
```diff
@@ -212,44 +213,39 @@ def reduce_pixel_shards(
...
         reducing_key (str): unique string for this task, used for resume files.
-        origin_pixel_numbers (list[int]): high order pixels, with object
-            data written to intermediate directories.
         destination_pixel_order (int): order of the final catalog pixel
         destination_pixel_number (int): pixel number at the above order
         destination_pixel_size (int): expected number of rows to write
             for the catalog's final pixel
         output_path (UPath): where to write the final catalog pixel data
+        ra_column (str): where to find right ascension data in the dataframe
+        dec_column (str): where to find declation in the dataframe
```
gitosaurus, line 235:
```suggestion
        dec_column (str): where to find declination in the dataframe
```

src/hats_import/catalog/arguments.py
```diff
@@ -45,6 +45,11 @@ class ImportArguments(RuntimeArguments):
     but the provided sorting will be used for any rows within the same higher-order pixel space."""
     add_healpix_29: bool = True
     """add the healpix-based hats spatial index field alongside the data"""
+    write_table_kwargs: dict | None = None
+    """additional keyword arguments to use when writing files to parquet (e.g. compression schemes)"""
```
gitosaurus, line 49:
```suggestion
    """additional keyword arguments to use when writing files to parquet (e.g. compression schemes)."""
```

gitosaurus (approved):
Looks good, but I recommend fixing the one typo before merging

PR #535, opened 2025-05-12
Remove directory-specific checks.

delucchi-cmu:
Checks against middle-directories are not useful on Google Cloud Storage file system, as these paths don't really exist. AWS pretends they do, but GCS does not. The checks that are failing aren't particularly useful, and the missing files/directories are caught elsewhere.


gitosaurus (approved):
LGTM!

PR #546, opened 2025-05-21
Run verification through typical `pipeline` construction

delucchi-cmu:
There are a lot of changes in here, that might not seem related, so I'll try to explain:

- adds some more imports in the top-level init. that's really not related, but if I send this Kostya, it will be ok
- don't use a client creation context for verification - there's no dask client needed (and no dask-client-creation arguments)
- don't use the try/except context for verification - there's no completion email address
- move the `write_mode` and `check_metadata` arguments into the `VerificationArguments` object, so there are no additional kwargs to pass to the verifier run method
- add tests to confirm that we send a completion email address.

### DIFF
--- a/src/hats_import/__init__.py
+++ b/src/hats_import/__init__.py
@@ -1,4 +1,10 @@
 """All modules for hats-import package"""
 
 from ._version import __version__
+from .catalog import ImportArguments
+from .collection import CollectionArguments
+from .index import IndexArguments
+from .margin_cache.margin_cache_arguments import MarginCacheArguments
+from .pipeline import pipeline, pipeline_with_client
 from .runtime_arguments import RuntimeArguments
+from .verification.arguments import VerificationArguments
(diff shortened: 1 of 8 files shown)


src/hats_import/__init__.py
hombit, line 1: Thank you!

src/hats_import/pipeline.py
hombit, line 1: Do we assume here that the verification pipeline would usually run fast? Or does it never use Dask?

tests/hats_import/test_pipeline_run.py
```diff
@@ -0,0 +1,59 @@
+import pandas as pd
+import pytest
+
+import hats_import
+from email.message import EmailMessage
+
+
+def test_runner(small_sky_object_catalog, tmp_path):
```
hombit, line 6: Nice!

PR #548, opened 2025-05-22
Refactor file_readers module into a subpackage

hombit:


tests/hats_import/catalog/test_file_readers.py
```diff
@@ -8,14 +8,14 @@
...
 from hats_import.catalog.file_readers import (
-    CsvPyarrowReader,
     CsvReader,
     FitsReader,
     IndexedCsvReader,
     IndexedParquetReader,
     ParquetReader,
     get_file_reader,
 )
+from hats_import.catalog.file_readers.csv import CsvPyarrowReader
```
delucchi-cmu, line 18: Why doesn't this one get imported at the `hats_import.catalog.file_readers` level too?

delucchi-cmu (approved):
Thank you! This is probably overdue.

tests/hats_import/catalog/test_file_readers.py
hombit: Fixed

PR #556, opened 2025-06-04
Be more numpy-ish with alignment handling

### LINKED ISSUE #538: Error when re-importing a nested catalog
**Bug report**

I'm not sure if it important that it is nested or not, but the code fails with the following error:

```
...

File ~/.virtualenvs/default/lib/python3.10/site-packages/hats_import/catalog/resume_plan.py:303, in <dictcomp>(.0)
    300     non_none_elements = alignment[alignment != np.array(None)]
    301     pixel_list = np.unique(non_none_elements)
    302     self.destination_pixel_map = {
--> 303         HealpixPixel(order, pix): count for (order, pix, count) in pixel_list if int(count) > 0
    304     }
    305 total_rows = sum(self.destination_pixel_map.values())
    306 if total_rows != expected_total_rows:

TypeError: cannot unpack non-iterable numpy.int32 object
```

The notebook is [here](https://github.com/lincc-frameworks/notebooks_lf/blob/main/lsdb/tns-x-ztf_alerts.ipynb), it takes few minutes to run it on Epyc

hats-import 0.5.2, hats 0.5.2

**Before submitting**
Please check the following:

delucchi-cmu:
Closes #538


dougbrn (approved):
LGTM, thank you!!!

Issue #569, opened 2025-07-11
Unclear error for a file corrupted on the reducing stage


hombit:
**Bug report**

Francisca Barra faced an issue with a large hats-import task (zubercal DR20): one of the output parquet files was corrupted, and importing failed on the finishing stage, because of the mis-matched number of rows. It took a long while to understand what was going wrong and which file was corrupted. It tuned to be that parquet dataset [here](https://github.com/astronomy-commons/hats/blob/b22733e04c492aaae1adc114f10bae3bcfb37f7d/src/hats/io/parquet_metadata.py#L59) misses that file. Maybe changing `exclude_invalid_files=True` to `False` would help

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [x] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

PR #577, opened 2025-08-11
Check for corrupted file if resuming reduce step.

delucchi-cmu:
Partial solution to #569. 

See also: https://github.com/astronomy-commons/hats/pull/539


src/hats_import/catalog/map_reduce.py
```diff
@@ -263,6 +263,14 @@ def reduce_pixel_shards(
         healpix_pixel = HealpixPixel(destination_pixel_order, destination_pixel_number)
         destination_file = paths.pixel_catalog_file(output_path, healpix_pixel, npix_suffix=npix_suffix)
         if destination_file.exists():
+            merged_table = pq.read_table(destination_file)
+            rows_written = len(merged_table)
```
hombit, line 267: 1. Do we actually need to read the full file, or the metadata would be enough?
2. Should we use `hats`' file reader for the cases of the remote `destination_file`?

hombit (approved):
Looks good, please see a couple of comments bellow

PR #582, opened 2025-09-04
Mimic reference API docs of nested-pandas

delucchi-cmu:
See similar change: https://github.com/astronomy-commons/lsdb/pull/853


delucchi-cmu:
Compare:
- preview: https://hats-import--582.org.readthedocs.build/en/582/reference.html
- current: https://hats-import.readthedocs.io/en/latest/autoapi/index.html

hombit (commented):
It looks really nice, but it misses everything but `*Arguments` classes

hombit:
@delucchi-cmu Let's add other stuff we mention in the docs, at least `pipeline*` functions and reader classes

delucchi-cmu:
Added sections for pipeline execution and file readers. Hierarchy makes it easier if we want to add even more things here in the future.

PR #584, opened 2025-09-04
Allow single file for input_file_list

### LINKED ISSUE #580: input_file_list should handle a single string
**Bug report**

If you pass a single string in the `input_file_list` argument, it will interpret as a list of characters.

We should either throw a warning when instance of string, or infer as a list with a single element.

**Before submitting**
Please check the following:

- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.

delucchi-cmu:
Let users pass in a single file for the `input_file_list` argument, and don't interpret is as a list of single characters.

Closes #580

### DIFF
--- a/src/hats_import/runtime_arguments.py
+++ b/src/hats_import/runtime_arguments.py
@@ -6,6 +6,7 @@ import re
 from dataclasses import dataclass
 from importlib.metadata import version
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
LGTM

PR #601, opened 2025-10-07
Add configuration for a healpix column name. 

delucchi-cmu:
Related to (but does not close) https://github.com/astronomy-commons/hats/issues/550

Requires https://github.com/astronomy-commons/hats/pull/560

- Adds the properties into the created catalog, when generating a `_healpix_29` column.
- Allows the use of a differently-named spatial index column in margin generation.
- Refreshes the unit test data, using the most recent properties.


delucchi-cmu:
> Looks good! I just have a quick question.
> 
> Since we added this custom healpix index logic to the margin generation, should we also allow creating object/source/index catalogs with a custom healpix index?

I'd rather not - this is something that we should support for catalogs that might not have this exact column name, but I'm fine with us being opinionated and always creating the spatial index column in the same way. If anyone requests us to name it something different, we could add that later, but I don't want to add that unless it becomes an issue.

PR #609, opened 2025-10-23
Explicit mention on input_file_list and number of workers.

delucchi-cmu:
This is a partial solution to problem "input_file_list, which is not always a list of files but a list of some strings, is confusing"

### DIFF
--- a/docs/catalogs/arguments.rst
+++ b/docs/catalogs/arguments.rst
@@ -122,9 +122,9 @@ There are a few ways to specify the files to read:
     in_file_paths.sort()
 
 .. important::
-    We will create one task per element inside the ``input_file_list``. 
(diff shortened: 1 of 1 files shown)


docs/catalogs/arguments.rst
```diff
@@ -121,6 +121,13 @@ There are a few ways to specify the files to read:
     in_file_paths = glob.glob("/data/object_and_source/object**.csv")
     in_file_paths.sort()
 
+.. important::
+    We will create one task per element inside the ``input_file_list``. 
+
+    If you have only one large file, we can only read the file with one worker.
```
hombit:
```suggestion
    If you have only one large file, we can only read the file with one Dask worker.
```

Issue #628, opened 2025-11-13
Option to not generate _metadata on catalog import


delucchi-cmu:
**Feature request**

This file is immediately deleted for PPDB. The hugging face writer testing would be nice to not write this file for now.

Requires change in `hats` package to add flag, similar to optionally creating thumbnail. Should default to True.

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #641, opened 2025-12-01
Make metadata creation optional

olivialynn:
Closes (with https://github.com/astronomy-commons/hats/pull/605) issue #628

*Another PR pair where unit tests here depend on the other PR getting merged—but CI passes locally, I pinky promise...*

**Description**:

The PR-pair adds a `create_metadata` flag to hats-import to make the creation of the `/dataset/_metadata` parquet file optional (but created by default).

This PR adds a `create_metadata` option to `ImportArguments`, which is then passed along to the call to hats' `write_parquet_metadata`.

**Question**: 

Do we want to add this flag to any of the other 3 places where we call `parquet_metadata.write_parquet_metadata` in the finishing stage in hats-import? These places are:
- `margin_cache.py` - I looked into this a little, and it seems that the metadata file is used immediately after creation to generate `margin_catalog_info`, to then make the property file ([L82 in margin_cache.pw](https://github.com/astronomy-commons/hats-import/blob/6ff0c90eb59cf9928f67e9fd7ad014e11ab34d52/src/hats_import/margin_cache/margin_cache.py#L82)). It should still be possible to get this done without a metadata file, but it was enough to give me pause and want to ask.
- `run_index.py`
- `run_conversion.py`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-hats @ git+https://github.com/astronomy-commons/hats.git@issue/hats-imp-628/optional-metadata
+hats @ git+https://github.com/astronomy-commons/hats.git@main
 nested-pandas @ git+https://github.com/lincc-frameworks/nested-pandas.git@main


olivialynn:
Build documentation is failing due to different installation order in workflow. Opened a PR in the PPT to rearrange this to match other workflows: https://github.com/lincc-frameworks/python-project-template/pull/557

Issue #529, opened 2025-05-05
source-object-nesting pipeline


delucchi-cmu:
Similar to the old SOAP (source-object-association-pipeline), a pipeline that attempts to match source and object partitions up to create nested light curves.

Pipeline should be kind of an intersection of catalog import and SOAP principles:
- do something similar to "soft soap" and find all source partitions with corresponding sources for all objects in the object partition
- for the object partition, nest the sources it contains
- if the object partition exceeds some threshold, split further
- write parquet files
- create catalog files.

delucchi-cmu:
There are some performance issues with my approaches, and I don't think this will be ready in the near-term.

PR #645, opened 2025-12-09
Move shared import arguments to parent class

delucchi-cmu:
Catalog arguments controlling the output parquet files are also useful for margin creation (and will be useful for the nested light curve pipeline).

Related to #529


src/hats_import/collection/arguments.py
```diff
@@ -208,6 +208,10 @@ def _get_subarg_dict(self):
         useful_kwargs = {
             "output_path": self.catalog_path,
             "addl_hats_properties": self.addl_hats_properties,
+            "npix_suffix": self.npix_suffix,
+            "npix_parquet_name": self.npix_parquet_name,
+            "write_table_kwargs": self.write_table_kwargs,
+            "row_group_kwargs": self.row_group_kwargs,
```
camposandro: Should we also pass the remaining parameters through?

```
skymap_alt_orders
should_write_skymap
create_thumbnail
create_metadata
```

The margin import only uses `create_metadata` but the rest should be ignored.

camposandro (approved):

PR #650, opened 2025-12-16
Use common hats utilities for size estimation and binning.

delucchi-cmu:
See https://github.com/astronomy-commons/hats/pull/613

And I promise this is all related to https://github.com/astronomy-commons/hats-import/issues/529

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-hats @ git+https://github.com/astronomy-commons/hats.git@delucchi/prefactor
+hats @ git+https://github.com/astronomy-commons/hats.git@main
 nested-pandas @ git+https://github.com/lincc-frameworks/nested-pandas.git@main


src/hats_import/catalog/map_reduce.py
```diff
@@ -113,9 +111,7 @@ def map_to_pixels(
     try:
         # Always generate the row-count histogram.
```
olivialynn:
```suggestion
```

src/hats_import/catalog/map_reduce.py
```diff
@@ -137,32 +133,24 @@ def map_to_pixels(
             use_healpix_29,
             read_columns,
         ):
-            # Always add to row_count histogram.
-            mapped_pixel, count_at_pixel = np.unique(mapped_pixels, return_counts=True)
-            row_count_histo.add(SparseHistogram(mapped_pixel, count_at_pixel, highest_order))
+            data_mem_sizes = None
 
             # If using bytewise/mem_size thresholding, also add to mem_size histogram.
```
olivialynn:
```suggestion
```

olivialynn (approved):
I like it! 

Suggesting to take out these two comments since the code they refer to is a little different now and they seem less relevant.

PR #652, opened 2025-12-17
Pipeline for nesting light curves onto object catalog.

### LINKED ISSUE #529: source-object-nesting pipeline

delucchi-cmu:
Initial implementation of a pipeline to create a new object catalog, based on original object catalog, plus a source catalog of individual rows per-detection. Detections are aggregated into a nested column, and the new catalog is created.

Closes #529


src/hats_import/nest_light_curves/arguments.py
```diff
@@ -0,0 +1,132 @@
...
+    to drop those from this list."""
+
+    ## Output - nested light curves
+    nested_column_name: str = "light_curve"
+    """The name for the group of nested columns."""
+    nested_sort_column: str = ""
+    """Field to sort the nested frame on. Typically timestamp."""
+    partition_strategy: str = "object_count"
+    partition_threshold: int = 50
+    highest_healpix_order: int = 3
```
hombit: this is a small default, should it be the same as for InputArguments?

src/hats_import/nest_light_curves/arguments.py
```diff
@@ -0,0 +1,132 @@
...
+    def read_source_columns(self):
+        """Determine list of columns to be read in the source catalog."""
+        if self.source_nested_columns is None:
+            # We intend to read all columns
+            return None
+        read_columns = list(set(self.source_nested_columns) | set([self.source_object_id_column]))
+        read_columns.sort()
+        return read_columns
+
+    def to_table_properties(self, total_rows=10, highest_order=4, moc_sky_fraction=22 / 7) -> TableProperties:
```
hombit: I probably miss the point here, but I don't understand any of these defaults...

hombit (commented):
Reviewed due to the Slack request, I'm not approving the PR because I believe @olivialynn is the primary reviewer.
Two major notes: 1) "inner" vs. "left" join, I believe that the right default is "left" until we believe that the object catalog is distributed independently, 2) it would be nice to check that all sources are used and nothing is lost in the inner/left join. I know that we have verification pipelines for this kind of stuff, but here we have access to these numbers, so we can check.

PR #659, opened 2026-01-06
Relax pixel_threshold limits

### LINKED ISSUE #658: Allow `pixel_threshold` bellow 100
**Bug report**

Working on [`mmu-hdf-to-hats`](https://github.com/UniverseTBD/mmu-hdf-to-hats), we found that some datasets have extremely large rows, hundreds of megabytes (see this 900MB [manga input file](https://users.flatironinstitute.org/~polymathic/data/MultimodalUniverse/v1/manga/manga/healpix=385/), which is a single object). However, `hats-import` fails when I set `pixel_threshold=1`:
```
ValueError: pixel_threshold should be between 100 and 1,000,000,000
``` 

**Environment Information**

hats-import v0.7.3

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, and any applicable data others will need to reproduce the problem.
- [x] I have included information about my environment, including the version of this package (e.g. `hats_import.__version__`)

delucchi-cmu:
Closes #658 

`pixel_threshold` must be a positive integer. Anything else is fine.

### DIFF
--- a/src/hats_import/catalog/arguments.py
+++ b/src/hats_import/catalog/arguments.py
@@ -119,8 +119,8 @@ class ImportArguments(RuntimeArguments):
             check_healpix_order_range(
                 self.lowest_healpix_order, "lowest_healpix_order", upper_bound=self.highest_healpix_order
             )
-            if not 100 <= self.pixel_threshold <= 1_000_000_000:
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
Looks good to me

PR #661, opened 2026-01-13
Use Path methods, where appropriate.

delucchi-cmu:
`is_regular_file` can be misleading, when pointing to an HTTP directory, and this is working as intended. Since the check is for finding a directory, `is_dir` is more appropriate.

The `is_regular_file` method is just a thin wrapper around path methods (as is `does_file_or_directory_exist`), so this PR uses those simple path methods, so the `file_io` methods can be removed later.


hombit (approved):
I have a concern about checking HTTP paths with `is_file/is_dir`, but it may go to a new issue because this PR doesn't change the previous behavior.

PR #662, opened 2026-01-16
Constrain dask versions

delucchi-cmu:
There is an odd regression in performance in newer versions of dask, but only under python=3.11. This allows all other versions of python to use newer versions of dask.


dougbrn (approved):
Looks good to me!

PR #673, opened 2026-03-26
Rename alignment argument, and shrink alignment, when possible.

delucchi-cmu:
## Change Description

Related to #529 

Implements changes to `generate_alignment` from https://github.com/astronomy-commons/hats/pull/653

I was looking for the warning introduced in the above hats PR, and went on a little side-quest to address other warnings.

## Code Quality
- [x] I have read the [Contribution Guide](https://hats-import.readthedocs.io/en/stable/guide/contributing.html) and [LINCC Frameworks Code of Conduct](https://lsstdiscoveryalliance.org/programs/lincc-frameworks/code-conduct/)

### DIFF
--- a/src/hats_import/catalog/resume_plan.py
+++ b/src/hats_import/catalog/resume_plan.py
@@ -69,6 +69,7 @@ class ResumePlan(PipelineResumePlan):
             if import_args.debug_stats_only:
                 run_stages = ["mapping", "finishing"]
             self.input_paths = import_args.input_paths
+            self.mapping_order = import_args.mapping_healpix_order
 
             # Set threshold_mode based on byte_pixel_threshold
             if hasattr(import_args, "byte_pixel_threshold") and import_args.byte_pixel_threshold is not None:
@@ -335,7 +336,7 @@ class ResumePlan(PipelineResumePlan):
         expected_total_rows,
         existing_pixels=None,
         raw_histogram_mem_size=None,
-    ) -> UPath:
+    ) -> tuple[UPath, int]:
         """Get a pointer to the existing alignment file for the pipeline, or
         generate a new alignment using provided arguments.
 
@@ -359,6 +360,7 @@ class ResumePlan(PipelineResumePlan):
             path to cached alignment file.
         """
         file_name = import_io.append_paths_to_pointer(self.tmp_path, self.ALIGNMENT_FILE)
+        self.mapping_order = hp.npix2order(len(raw_histogram))
         if not file_name.exists():
             # If existing_pixels, create an incremental alignment.
             if existing_pixels:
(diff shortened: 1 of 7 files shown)


src/hats_import/catalog/resume_plan.py
```diff
@@ -359,6 +360,7 @@ def get_alignment_file(
             path to cached alignment file.
```
olivialynn, line 360: I don't remember the formatting for a google doc tuple return, but maybe providing the following info:
tuple of Upath, int
Where UPath is the path to the cached alignment file, and int is the mapping_order (the healpix order that was used when mapping)

src/hats_import/catalog/resume_plan.py
```diff
@@ -400,14 +403,15 @@ def get_alignment_file(
                 for (order, pix, row_count) in pixel_list
                 if int(row_count) > 0
             }
+            self.mapping_order = hp.npix2order(len(alignment))
```
olivialynn, line 406: I think I may be missing something - why do we set `self.mapping_order` above on L363 and don't seem to use it, only to set it again here (if I'm correct that this conditional will always be True on a fresh run, where we don't have a `self.destination_pixel_map` set yet)?

src/hats_import/catalog/resume_plan.py
```diff
@@ -69,6 +69,7 @@ def __init__(
             if import_args.debug_stats_only:
                 run_stages = ["mapping", "finishing"]
             self.input_paths = import_args.input_paths
+            self.mapping_order = import_args.mapping_healpix_order
```
olivialynn, line 72: Do we ever use this before overwriting it in get_alignment_file (L363)?

olivialynn (commented):
I like it! 

Does this plan to fully phase out the user-set `mapping_healpix_order`? If so, should we also remove it from the hats_import arguments.py (https://github.com/astronomy-commons/hats-import/blob/main/src/hats_import/catalog/arguments.py#L86)? You can see in one of my comments in resume_plan that I'm struggling to fully follow mapping_order through the code, but hoping it's just a case of monday brain

src/hats_import/catalog/resume_plan.py
delucchi-cmu: You're correct - we don't need to keep this as a member anymore.

src/hats_import/catalog/resume_plan.py
delucchi-cmu: Fixed - thanks!

delucchi-cmu:
> I like it!
> 
> Does this plan to fully phase out the user-set `mapping_healpix_order`? If so, should we also remove it from the hats_import arguments.py (https://github.com/astronomy-commons/hats-import/blob/main/src/hats_import/catalog/arguments.py#L86)? You can see in one of my comments in resume_plan that I'm struggling to fully follow mapping_order through the code, but hoping it's just a case of monday brain

There is no plan to phase out the `mapping_healpix_order`. This is still a useful first approximation at the order that is later refined.

- At the mapping stage, we need to have some reasonable maximum order at which to build the histograms. Always doing this at order 20 is costly in histogram size, and is rarely needed to be greater than 10.
- At the binning stage, we can refine that maximum order, given the full list of pixels in the final partitioning (which is what's added by this PR).
- At the splitting stage, we need to know what order the alignment is in. It doesn't need to be the same order as was used in the mapping, since these are independent stages.

olivialynn (approved):
Looks good!

PR #698, opened 2026-07-10
Generate additional static files. Test against full dependencies.

### LINKED ISSUE #697: Add flags for creating new static files
**Feature request**

- skymap.png
- partition_info.png
- README.md
- index.html

delucchi-cmu:
## Change Description

Will require https://github.com/astronomy-commons/hats/pull/704 for tests.

Closes #697 

Adds matplotlib as an optional dependency (to generate those PNGs). This PR does some optional/full dependency testing management.


.github/workflows/smoke-test.yml
```diff
@@ -38,6 +38,15 @@ jobs:
     - name: Run unit tests with pytest
       run: |
         python -m pytest tests
+    - name: Install FULL dependencies
+      run: |
+        uv pip install --system -e .[full]
+    - name: List FULL dependencies
+      run: |
+        pip list
```
MayoKashyap, line 52: What does this code block exactly do..? I'm a little lost on the file name too.

src/hats_import/collection/run_import.py
```diff
@@ -34,8 +36,25 @@ def run(args, client):
...
+        step_progress.update(1)
+        if args.create_partition_info_png:
+            write_partition_info_png(args.catalog_path)
+        step_progress.update(1)
+        if args.create_summary_html:
+            write_catalog_summary_file(args.catalog_path, fmt="html")
+        step_progress.update(1)
+        if args.create_summary_md:
+            write_catalog_summary_file(args.catalog_path, fmt="markdown")
+        step_progress.update(1)
```
MayoKashyap, line 57: I'm assuming step_progress.update(1) has something to do with a progress bar...but why would we ned it here?

src/hats_import/collection/run_import.py
```diff
@@ -34,8 +36,25 @@ def run(args, client):
         simple_progress_bar=args.simple_progress_bar,
         tqdm_kwargs=args.tqdm_kwargs,
     ) as step_progress:
+        # pylint: disable=duplicate-code
```
MayoKashyap, line 39: Did you have to add this because you wrote the exact same code in run_import.py

pyproject.toml
```diff
@@ -47,6 +47,7 @@ full = [
     "ipykernel", # Support for Jupyter notebooks
     "ipywidgets", # useful for tqdm in notebooks.
     "healpy", # used only in hipscat conversion
+    "matplotlib", # used when creating skymap.png or partition_info.png
```
MayoKashyap, line 50: im assuming this line adds matplotlib as an optional dependency...

MayoKashyap (approved):

.github/workflows/smoke-test.yml
delucchi-cmu: I've added some implementation comments to the file.

The gist is that this runs a nightly job in a github workflow. This serves to test that *upstream* changes don't break our code (e.g. if numpy changes the behavior of a sort, we'll know the next day instead of the next time someone opens a pull request).

We have a handful of optional dependencies that only need to be installed when you're using those particular features. Several parties have expressed a desire to keep the required dependencies to a minimum. We still want to make sure that for those users who want the full set of features, we are testing the functionality. So this file splits the testing into "just the required dependencies" and "full dependencies".  This kind of split *should* have already been present, since this is not the first optional dependency we've added. But better late than never...

src/hats_import/collection/run_import.py
delucchi-cmu: Yes! There's a lot of duplicated code between the import pipelines, particularly for the finishing stage. The pylint warning can be helpful for suggesting places where the code could be abstracted and shared, but this is a known place where the code just needs to be fairly similar.

src/hats_import/collection/run_import.py
delucchi-cmu: Yes! We have progress bars all over the place in the import pipeline, but for most of them, they are based on some iterable (e.g. for each input file, or for each output parquet file). The "finishing" stage is a little special, because there are a fixed number of files to write, but it has still been useful to have this as a progress bar.

tqdm still supports this use case with manual control over the bar with the `step_progress.update()` call.

pyproject.toml
delucchi-cmu: Yes - the GitHub UI is cutting off a bunch of the config above this that shows it's in the `[project.optional-dependencies]` > `full` block.

`matplotlib` is a HUGE dependency, and we would not want every hats-import user to have to install it, even if they don't want to generate any plots.

delucchi-cmu (commented):
Great questions - and your assumptions are all correct!

PR #715, opened 2026-08-10
Add warnings and use parquet kwargs for IndexArguments

### LINKED ISSUE #708: `IndexArguments` silently ignores arguments
**Bug report**

`hats_import.index.arguments.IndexArguments` is [documented](https://hats-import.readthedocs.io/en/latest/reference/api/hats_import.IndexArguments.html) as accepting the following args, and passing them in does not raise an error or warning, but the args are ignored.

- write_table_kwargs
- npix_suffix
- npix_parquet_name
- create_per_partition_stats

<details>
<summary>Details</summary>

Pseudocode I used with those args to create an Index Table:

```python
hats_import.index.arguments.IndexArguments(
    input_catalog_path=<input-path>,
    output_path=<output-path>,
    output_artifact_name=<artifact-name>,
    indexing_column="prefname",
    drop_duplicates=False,
    create_per_partition_stats=True,
    npix_suffix="/",
    npix_parquet_name="part0.snappy.parquet",
    extra_columns=[<extra-col-1>, <extra-col-2>],
    write_table_kwargs={
        "compression": "snappy",
        "write_statistics": True,
        # Using v1.0 for compatibility. See https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html.
        "version": "1.0",  
    },
    tmp_dir=<tmp-dir>,
    division_hints=[<division_hints>],
)

with dask.distributed.Client(
    n_workers=<n_workers>,
    threads_per_worker=1,
    memory_limit=None,
    local_directory=<local-dir>,
) as client:
    hats_import.pipeline_with_client(args, client)
```

Showing that `write_table_kwargs` is ignored:

```python
>>> md = pyarrow.parquet.read_metadata(<path-to-index-file.parquet>)
>>> md
<pyarrow._parquet.FileMetaData object at 0x7f64e66ca430>
  created_by: parquet-cpp-arrow version 23.0.0
  num_columns: 28
  num_rows: 1769526
  num_row_groups: 2
  format_version: 2.6  # <--- if write_table_kwargs was used, this would be 'format_version: 1.0'
  serialized_size: 8434

>>> md.row_group(0).column(0).compression
'SNAPPY'  # <--- as I wanted, but had I wanted something else it would still be snappy.
```

Looking at the code in `src/hats_import/index/map_reduce.py`, it appears that hats-import doesn't pass any kwargs when writing Index Table files, so pyarrow defaults are always used.

</details>

After trying it and seeing those were ignored, I realized that the last three don't make sense for an Index Table, but it took my tired brain a minute 😬 . The docs could state that they're ignored in this case.

`write_table_kwargs` still seems like it should work here, but if there's some reason it can't then just documenting that would be helpful. I wanted to use it to set `compression="snappy"` (since I know you typically default to ZSTD) and `version="1.0"` ([for compatibility](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.write_table.html)).

**Environment Information**

hats-import v0.10.2

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, and any applicable data others will need to reproduce the problem.
- [x] I have included information about my environment, including the version of this package (e.g. `hats_import.__version__`)

smcguire-cmu:
Closes #708


camposandro (approved):
Looks great!

